PR #537, opened 2018-05-07
Update adaptivecards to 1.0.0

mattmazzola:


src/components/modals/AdaptiveCardViewer/AdaptiveCardViewer.tsx
```diff
@@ -9,23 +9,21 @@ import { connect } from 'react-redux';
...
-    renderOptions(): any {
-        return {
-            hostConfig: hostconfig
-        }
+    getAdaptiveCardHostConfig(): any {
+        return new AdaptiveCards.HostConfig(adaptiveCardHostConfig)
     }
 
     getTemplate(): any {
-        let templateString = JSON.stringify(this.props.template.body);
```
mattmazzola: This was the only questionable part. This looks like a mistake that the old renderer just happened to accept.  The template.body is already a string, so calling JSON.stringify on it escaped it further which mean the subsequent JSON.parse would only turn it back into a string instead of actual JSON which resulted in render failure.

PR #546, opened 2018-05-10
Vscolor

LarsLiden:


.vscode/settings.json
```diff
@@ -17,5 +17,8 @@
         "selectable",
         "to",
         "truthy"
-    ]
+    ],
+    "workbench.colorCustomizations": {
+        "activityBar.background": "#164e86"
```
mattmazzola: Is there a way to make this a user setting so it doesn't effect everyone? The work place settings are intended for the project which are shared.

LarsLiden: Talked to them about this, an user settings go across all projects, so the only way to differentiate is on workspace

mattmazzola: I think there is a way to inherit settings from parent folder but not be global.

Maybe: https://code.visualstudio.com/docs/editor/multi-root-workspaces

Otherwise every other develop would have to stash / unstash changes to avoid this color

mattmazzola:
Ah well, already merged.  I don't use activity bar so I don't think it will affect me.

PR #555, opened 2018-05-16
1343 Remove datetime pre-built entity (deprecated; use datetimeV2 ins…

LarsLiden:
…tead)

Looks like only some of the languages are fully migrated in LUIS.  FR for example now has DateTimeV2, Italian has Datetime.  As we are currently only supporting English, I'm just updating that.  We can update the others when we add those languages (as we'll have to do that anyway to see what state they are in)


LarsLiden:
close and re-open to trigger PR build (stuck)

PR #581, opened 2018-06-01
 bug(actionscorer): Fix sorting algorithm

mattmazzola:
- Use localeCompare for strings
- Differentiate between NotScoreable but still usable and NotScoreable and not usable.

### DIFF
--- a/src/components/modals/ActionScorer.tsx
+++ b/src/components/modals/ActionScorer.tsx
@@ -28,6 +28,7 @@ const ACTION_BUTTON = 'action_button'
 const MISSING_ACTION = 'missing_action'
 
 interface IRenderableColumn extends OF.IColumn {
+    getSortValue: (action: ScoredBase, component: ActionScorer) => number | string
     render: (x: ScoredBase, component: ActionScorer, index: number) => React.ReactNode
 }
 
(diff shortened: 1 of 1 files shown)


src/components/modals/ActionScorer.tsx
```diff
@@ -125,6 +128,30 @@ function getColumns(intl: InjectedIntl, hideScore: boolean): IRenderableColumn[]
...
+                // TODO: Fix type so we can use typed property access
+                const score = scoredBase['score']
+
+                // If score base does not have score it's either not scorable or not available
+                // prioritize not scorable over not available but both at bottom of list
+                if (!score) {
+                    if (scoredBase['reason'] === ScoreReason.NotAvailable) {
+                        return -100;
+                    } else {
+                        let isAvailable = component.isUnscoredActionAvailable(scoredBase as UnscoredAction);
```
mattmazzola: This is the part that is really weird to me.

The server is returning objects that have `reason: notScorable` and also `reason: notAvailable` and yet even for actions that are 'notScorable' we classify them as "available" or "not available" which overlaps with the actual `notAvailable` from service.

LarsLiden: A "notScorable" action may be available or not - the server is running an older model (i.e. training still taking place) so it doesn't know.  In that case the client does the calculation to determine which category it is in

PR #584, opened 2018-06-05
fix: Update tipComponents to pass through other properties intended for main element

mattmazzola:
- Can now pass attributes like id or data-testid from TC.Dropdown to the internal OF.Dropdown
- Example is shown in ActionCreatorEditor


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -803,6 +803,7 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                 <div className="cl-modal_body">
                     <div>
                         <TC.Dropdown
+                            data-testid="dropdown-action-type"
```
mattmazzola: Example of how to add `data-testid`

PR #594, opened 2018-06-06
Minervaen/helloworldtest

MinervaEn:
- Adding hello world test.
-  fixing duplicated data-testid name under actioncreatoreditor modal.

### DIFF
--- a/cypress/integration/helloworld_test.ts
+++ b/cypress/integration/helloworld_test.ts
@@ -1,0 +1,122 @@
+describe('Hello world e2e', function () {
+    const newAppName = `e2e-HelloWorld-${new Date().getTime() - 1528151000000}`
+    const username = `UserName`
+    const action01 = 'Hello World!'
+    /**
+     * FEATURE: New Model
+     */
+    it('CREATE NEW MODEL with random name and verify name on application page', function () {
+      // Open application
+      cy.visit('http://localhost:5050')
+  
+      // Click the button to create app
+      cy
+        .get('[data-testid="apps-list-button-create-new"]')
+        .click()
+  
+      // Ensure that name input is focused
+      cy.focused()
+  
+      cy
+        .get('[data-testid="app-create-input-name"]')
+        .type(newAppName)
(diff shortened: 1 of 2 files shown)


cypress/integration/helloworld_test.ts
```diff
@@ -0,0 +1,122 @@
...
+        .get('a[href$="/actions"]')
+        .click()
+  
+      // Click new action button
+      cy
+        .get('[data-testid="actions-button-create"]')
+        .click()
+  
+      // Click Action Type dropdown
+      cy
```
mattmazzola: This seems unnecessary. It re-selects what was already selected.

cypress/integration/helloworld_test.ts
```diff
@@ -0,0 +1,122 @@
...
+        .get('[data-testid="app-index-title"]')
+        .should(el => {
+          expect(el).to.contain(newAppName)
+        })
+    })
+  
+    /**
+     * FEATURE:  New Identity
+     */
+    it('it should create a NEW ENTITY', () => {
```
mattmazzola: `it('it` is redundant. Remove the second one that's part of test description

cypress/integration/helloworld_test.ts
```diff
@@ -0,0 +1,122 @@
...
+  
+      // Click Action Type dropdown
+      cy
+        .get('[data-testid="dropdown-action-type"]')
+        .click()
+        .click()
+  
+      // Enter action response phrase: Hello World!
+      cy
+        .get('div[data-slate-editor="true"]')
```
mattmazzola: This should be more specific to ensure you get the correct element in cases where the are multiple.  Also, would prefer not to couple selector to implementation detail of slate.  In other words, do not rely on that attribute.  Should only know about `contenteditable`

MinervaEn:
@mattmazzola thanks for your PR review, I will address you comments and update the test file accordingly. Also I will adjust to and follow the team's contribution's practices.

PR #607, opened 2018-06-14
Minervaen/logs

MinervaEn:
- Added test step for Log Dialogs 
- Updated file names to match other source files
- Renamed file to match file functionallity

PR #608, opened 2018-06-15
test: Adding Log Dialogs test step to hello world suite

MinervaEn:
- this PR adds a new test step to validate the creation of Log Dialogs
- adds the corresponding data-testids


mattmazzola:
It looks like this is similar to previous PR #607 but just changed data test ids.  I assume you didn't intentionally try to circumvent the review process, but please don't do this in the future.  Simply update the existing PR. As it is, I can't relate your code changes to the comments I previously made to see if they were resolved.

cypress/support/components/homepage.js
```diff
@@ -39,10 +40,7 @@ function createNewModel(modelName) {
 
 /** Delete an existent Model */
 function deleteModel(modelName) {
-  cy.server()
-  cy.route('GET', '/apps?**').as('getHomePage')
-  cy.visit('http://localhost:5050')
-  cy.wait('@getHomePage')
+  navigateTo()
```
mattmazzola: I don't think it's a good practice to nest these functions.  The test should call `navigateTo` and then `deleteModel`

PR #615, opened 2018-06-19
test: error handling updates

MinervaEn:
- removed individual listeners and added a before each listener instead
- added a utils testlog file to standardize the logging to console.
- added verbose to console logs.


cypress/support/utils/testlog.js
```diff
@@ -0,0 +1,34 @@
...
+function printResult(result) {
+    console.log(`${now()} - Test ${result.state}...`);
+}
+
+function printStep(description) {
+    console.log(`${now()} ====>> ${description}...`);
+}
+
+function printDoubleLine(length) {
+    console.log(`${now} ${"*".repeat(length)}`);
```
mattmazzola: Looks like a typo: `${now}` -> `${now()}`

cypress/support/utils/testlog.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function testHeader(testName) {
+    console.log(`${now()} ${"*".repeat(60)}`);
```
mattmazzola: Re-use the value of `now()` rather than recomputing

cypress/support/utils/testlog.js
```diff
@@ -0,0 +1,34 @@
...
+    console.log(`${now()} - ${"-".repeat(30)}>`)
+    console.log(`${now()} - ${err.stack} `);
+    console.log(`${now()} - <${"-".repeat(30)}`)
+}
+
+function now() {
+    return Cypress.moment().format("MMMDD-HH:mm.ss");
+}
+
+export { now, printDoubleLine, printResult, testHeader, reportError, printStep }
```
mattmazzola: Keep consistent naming.  This file is should be a logger, and re-use same javascript vocabulary as `log` instead of `print`.  In the web printing refers to printing on paper or to a device so I would prefer to keep the prefix `log`.

Also, instead of `reportError` I would prefer to use the builtin `console.error` or `console.warn`  I think they already print stack traces

MinervaEn:
All green! 
-removed individual listeners and added a before each listener instead
-added a utils testlog file to standardize the logging to console.
-added verbose to console logs.
-renamed function names to use  the prefix log instead of print.

PR #617, opened 2018-06-20
test: enabling debugging output

MinervaEn:
- using DEBUG to output cypress proxy and reporter logs.


MinervaEn:
this change is going nowhere. So, closing for now.

PR #620, opened 2018-06-25
increasing requestTimeout to allow requests to complete

MinervaEn:
It seems like there are some requests that take longer than 5 seconds to resolve (cypress's default value), which end up on intermittent failures.
This change will allow cypress to wait for the requests to complete before proceeding with next validations/steps.


mattmazzola:
This looks like duplicate of #621

PR #632, opened 2018-06-27
fix: add aria-live area to announce Retrying when Retry is clicked

mattmazzola:
Fixes VSTS#1383


src/routes/Apps/App/Dashboard.css
```diff
@@ -0,0 +1,4 @@
+.cl-dashboard-retry {
```
LarsLiden: Any reason not to re-use .cl-screen-reader?

mattmazzola: Yes, `.cl-screen-reader` came after this PR. It was not available to use.

PR #635, opened 2018-06-27
 fix: spacing incorrect on help icons

LarsLiden:


mattmazzola:
This look like it's trying to apply a fix to the label instead of only the icon.

LarsLiden:
I did both.  Fixed the icon but thought the spacing was still somewhat ambiguous so I tightened it up

mattmazzola:
That's not something we should adjust.  It's part of the framework.  A label and input are a unit.  It would be more appropriate to adjust the spacing between input above which is separate component to fix the issue.

As it is now it's too close and now is inconsistent with other labels:
image

mattmazzola:
Example of spacing we use in other places which makes it more clear which input the label is associated with:
image

I'm in favor of this, but whichever we choose I really just want it all to be consistent and since some designer already decided the appropriate label distance / whitespace I think we should respect that and not diverge.

LarsLiden:
I find the top left image to be confusing as the text fall exactly between the two input boxes.  Agree thought that it should be consistent across the board.   Perhaps better to overwrite the base css then to add an additional css class

mattmazzola:
I think you are referring to the image showing inconsistent label heights, not the one showing the separation between inputs.  The Search boxes seem very clear.

LarsLiden:
This is how it looks with and without the change.   I find the one on the right to be less confusing:

image  image

mattmazzola:
Yea, that's why i'm asking you to compare it with other image showing better spacing between inputs, not those two which are both bad.

LarsLiden:
I see, you are suggesting we increase the spacing between input boxes rather than decrease the spacing between label and input box.   I lean towards the latter as it means the user is less likely to need to scroll to see all the input boxes on the Action creator.

PR #643, opened 2018-06-29
test: checking in timeout changes 

MinervaEn:
trying to check in again.
- test files have been removed
- and timeouts have been set prior each test starts (as workaround to cypress issue with setting global timeouts).


MinervaEn:
We have discussed this fix many times, so checking in.

PR #678, opened 2018-07-27
fix: enable strict null checks

mattmazzola:
This will help catch logical errors and make it more clear when we should refactor certain code to remove excessive null checks


LarsLiden:
Looks like bbnpm leaked back in on this checkin...

mattmazzola:
Ah, damn. Thanks will fix

PR #706, opened 2018-09-18
fix: renamed traindialogmodal.js to editdialogmodal.js

SDESkowronski:
This is a change to test code, not product code. Fixes broken test runs due to recent product code changes.


LarsLiden:
You should also delete logdialogmodel.js as it doesn't exist any more

PR #722, opened 2018-10-12
Typo on string #42

kant:
* `onversationlearner` or `conversationlearner` on paragraphs #03 & 07?
* Plus minor fixes proposals


LarsLiden:
Thanks, Dario!

PR #740, opened 2018-10-25
Refactor Test Code & Folder Structure for Test Runner Menu

SDESkowronski:
The code was refactored so that we can better control the Cypress Test Runner's menu of test cases and include an ALL option for each test category and also for ALL categories.


LarsLiden:
@SDESkowronski  - you seem to be checking in a bunch of changes all at once.  Would prefer that you create a new branch for each check-in so that the PRs are small.  It's difficult to code review when the change set is so large

PR #757, opened 2018-10-31
fix: bug 1495 add training status to traindialog

pnwbrian:
Add training status container to traindialog, and flex inline


LarsLiden:
Also needs to be added to EditDialogAdmin

PR #764, opened 2018-11-02
fix: bug 1541 - fix model date created sorting

pnwbrian:
Switch from non-unique portion of the timestamp to fully qualified and expanded YYYYMMDDHHmmSS


src/components/ExtractorResponseEditor/EntityPicker.css
```diff
@@ -52,7 +52,7 @@
     list-style: none;
     padding: 0;
     padding-bottom: 0.25em;
-    border-bottom: 1px solid lightgrey;
```
LarsLiden: What does this change do?

PR #770, opened 2018-11-05
fix: bug 1575: push close button down to panel gutter

pnwbrian:
Introduced new fixed location gutter class for UX button lolcation consistency


LarsLiden:
built timed out.  re-trigger

PR #775, opened 2018-11-07
1105 bug bash.b

pnwbrian:
ActionScorer UX fixes from 11/05 bug bash; and new localized string helper

### DIFF
--- a/src/components/modals/ActionScorer.tsx
+++ b/src/components/modals/ActionScorer.tsx
@@ -45,8 +45,12 @@ function getColumns(intl: InjectedIntl, hideScore: boolean): IRenderableColumn[]
             render: (action, component, index) => {
 
                 const selected = (component.props.dialogType !== DialogType.TEACH && index === 0)
-                const buttonText = selected ? 'Selected' : 'Select'
-                if (!component.props.canEdit) {
+                const buttonText = intl.formatMessage({
+                    id:selected ? FM.BUTTON_SELECTED : FM.BUTTON_SELECT,
+                    defaultMessage:selected ? Util.getDefaultText(FM.BUTTON_SAVE_EDIT) : Util.getDefaultText(FM.BUTTON_SAVE_EDIT)
+                    });
+
+                    if (!component.props.canEdit) {
                     return (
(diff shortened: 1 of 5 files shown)


src/components/modals/ActionScorer.tsx
```diff
@@ -45,8 +47,12 @@ function getColumns(intl: InjectedIntl, hideScore: boolean): IRenderableColumn[]
             render: (action, component, index) => {
 
                 const selected = (component.props.dialogType !== DialogType.TEACH && index === 0)
-                const buttonText = selected ? 'Selected' : 'Select'
-                if (!component.props.canEdit) {
+                const buttonText = intl.formatMessage({
+                    id:selected ? FM.BUTTON_SELECTED : FM.BUTTON_SELECT,
+                    defaultMessage:selected ? Util.getDefaultText(FM.BUTTON_SAVE_EDIT) : Util.getDefaultText(FM.BUTTON_SAVE_EDIT)
```
LarsLiden: This looks like a bug - both at BUTTON_SAVE_EDIT ?

src/components/modals/ActionScorer.tsx
```diff
@@ -275,14 +293,17 @@ class ActionScorer extends React.Component<Props, ComponentState> {
...
+            cardViewerShowOriginal: false,
+            isAlreadySelectedOpen: false
         };
         this.handleActionSelection = this.handleActionSelection.bind(this);
         this.handleDefaultSelection = this.handleDefaultSelection.bind(this);
         this.handleOpenActionModal = this.handleOpenActionModal.bind(this);
         this.renderItemColumn = this.renderItemColumn.bind(this);
         this.onColumnClick = this.onColumnClick.bind(this);
         this.focusPrimaryButton = this.focusPrimaryButton.bind(this);
+        this.showPopup = this.showPopup.bind(this);
```
LarsLiden: Would prefer var names to be more detailed in case we add other one in the future (i.e. "showAlreadySelectedPopup, etc

pnwbrian:
Investigating higher level logic errors in this code block found during peer review

PR #782, opened 2018-11-08
fix: bug 1593; scroll bar hides status text

pnwbrian:
Status area now includes a little right side margin bumper. Bumped up the area's allocation a bit while here


pnwbrian:
Regression fix: header needs to be always visible

PR #786, opened 2018-11-19
fix: bug - 1597 - smarter abandonment confirmation prompt

pnwbrian:
Hoist existing Save button logic for greater component reach, and tap existing close modal logic

### DIFF
--- a/src/components/modals/TeachSessionModal.tsx
+++ b/src/components/modals/TeachSessionModal.tsx
@@ -46,7 +46,8 @@ interface ComponentState {
     // If activity selected its index
     selectedActivityIndex: number | null,
     // If activity was part of existing history, the actual item
-    selectedHistoryActivity: Activity | null
+    selectedHistoryActivity: Activity | null,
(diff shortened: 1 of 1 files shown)


src/components/modals/TeachSessionModal.tsx
```diff
@@ -673,6 +679,7 @@ class TeachModal extends React.Component<Props, ComponentState> {
         let saveDisable = !this.state.hasTerminalAction 
                         || this.props.teachSession.dialogMode === CLM.DialogMode.Extractor
                         || this.props.teachSession.botAPIError !== null
+        this.state.promptOnAbandon = !saveDisable
```
LarsLiden: This doesn't seem to be fixing what the bug is - namely don't ask abandon if there is no data yet.  This seems to remove the prompt for other unrelated reasons (i.e. in extractor mode) which doesn't make sense.  Let's talk

pnwbrian:
Reworking dialog logic

PR #790, opened 2018-11-19
fix: bug 1597 - smarter abandonment confirmation prompt

pnwbrian:
Quietly close if no training provided instead of prompting to confirm


pnwbrian:
Rerunning

PR #791, opened 2018-11-19
feat: add resolver type to custom entities

msft-shahins:


src/react-intl-messages.ts
```diff
@@ -166,6 +166,7 @@ export enum FM {
     ENTITIES_CREATEBUTTONTEXT = 'Entities.createButtonText',
```
LarsLiden: Can FM.ENTITYCREATOREDITOR_FIELDS_ALWAYSTAG_LABEL be deleted from here?

src/components/ToolTips/ToolTips.tsx
```diff
@@ -45,6 +45,7 @@ export enum TipType {
     ENTITY_ALWAYS_EXTRACT = 'alwaysExtract',
```
LarsLiden: I believe this can be deleted too...

PR #801, opened 2018-11-21
fix: bugs 1629,1625 - GRAMMAR - reword duplicate entity popup, model …

pnwbrian:
reworded confirmation prompts, and wired up new getDefaultText helper


src/react-intl-messages.ts
```diff
@@ -877,7 +877,7 @@ export default {
         [FM.ERROR_PRIMARYBUTTON_TEXT]: 'Ok',
 
         // ExtractConflictModal
-        [FM.EXTRACTCONFLICTMODAL_TITLE]: 'Entity labelling conflicts with an existing labelling',
+        [FM.EXTRACTCONFLICTMODAL_TITLE]: 'Entity label already in use',
```
LarsLiden: Wording change still isn't right.  The issue isn't that the label is in use, it's that it's being labelled differently than in a previous example (i.e. labelled as "numTables" in one example as "numPeople" in another, or labelled as "animal" in one example an NOT LABELLED in another.)   Open to ideas?   Probably also need a tool tip for this

pnwbrian: i'll chew on this one, and agree on the tooltip

PR #806, opened 2018-11-22
fix: Fix delete all models and CreateModels.AllEntityTypes

SDESkowronski:
Fixed a bug I noticed in delete all models that was due to a prior code change I had made.

Fixed a bug in CreateModels.AllEntityTypes that was due to a change in product code.

Fixed the test runner changes I just made in my last PR due to wanting to keep DeleteAllModels out of the main list of tests and instead keeping it as a separate tool.


.vscode/settings.json
```diff
@@ -21,6 +21,5 @@
     ],
     "workbench.colorCustomizations": {
         "activityBar.background": "#164e86"
-    },
-    "editor.formatOnSave": true
```
SDESkowronski: This was causing vscode to change the formatting of my test code files. I really don't want a robot changing the format of my code. While in many cases that is okay, it is not okay in all cases. So I'd rather be responsible for the format myself.

pnwbrian: While I agree, the team collectively is moving to the vscode styled formatting (aka typescript's)...

cypress/support/components/HomePage.js
```diff
@@ -13,19 +13,14 @@ export function ClickSubmitButton()             { cy.Get('[data-testid="model-cr
 export function UploadImportModelFile(name)     { cy.UploadFile(name, `[data-testid=model-creator-import-file-picker] > div > input[type="file"]`)}
 
 export function ClickDeleteModelButton(row)     { cy.Get(`[data-list-index="${row}"] > .ms-FocusZone > .ms-DetailsRow-fields`).find('i[data-icon-name="Delete"]').Click() }
-export function ClickConfirmButton()            { cy.Get('.ms-Dialog-main').contains('Confirm').Click() }
+export function ClickConfirmButton()            { return cy.Get('.ms-Dialog-main').contains('Confirm').Click() }
```
pnwbrian: Does Click() really return a boolean?

SDESkowronski: No, it does not return a Boolean. It returns the promise that allows us to chain the command with a .then() call which allows our code to execute in the proper time frame. This change was necessary and correct.

PR #818, opened 2018-11-28
fix: bug 1647 luis deep link changed

LarsLiden:


LarsLiden:
Re-triggering.  Travis build failed to report back to Github

PR #827, opened 2018-11-28
Rename Models and Fix Delete All to Delete All Test Generated Models

SDESkowronski:
- Test generated models have new names.
- All test generated models to begin with "z-".
- Now only the test generated models will be deleted.


cypress/support/components/HomePage.js
```diff
@@ -18,7 +20,28 @@ export function ClickConfirmButton()            { return cy.Get('.ms-Dialog-main
 export function GetModelListRowCount() 
 {
   return cy.Get('[data-automationid="DetailsList"] > [role="grid"]')
-  .then((gridElement) => { var rowCount = +gridElement.attr('aria-rowcount') -1; return rowCount })
+  .then(gridElement => { var rowCount = +gridElement.attr('aria-rowcount') -1; return rowCount })
```
SDESkowronski: Brian, this is what I was talking about. I can get this 'return rowCount' statement to work and actually return the value, but in the method right below it, DeleteNextTestGeneratedModel(), I could not get the code working that way. I had to create my own promise to get it to work.

PR #865, opened 2018-12-10
feat: add close button to help panel (Problem #3)

mattmazzola:
I had the button placed absolutely positioned to stay on the bottom since the content of the modal shrinks to smallest height possible.

The problem is that text can go behind the button and broken.  I added a white padding to try to keep these a little separate.  An alternative would be to reserve space at the bottom so nothing can crosss the bottom but it takes up unnecessary space in certain scenarios

### DIFF
--- a/src/components/HelpPanel.tsx
+++ b/src/components/HelpPanel.tsx
@@ -10,6 +10,9 @@ import { connect } from 'react-redux';
 import * as ToolTip from './ToolTips';
 import * as OF from 'office-ui-fabric-react';
 import { setTipType } from '../actions/displayActions'
+import { injectIntl, InjectedIntlProps } from 'react-intl'
+import { FM } from '../react-intl-messages'
+import './HelpPanel.css'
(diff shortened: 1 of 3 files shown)


src/components/HelpPanel.tsx
```diff
@@ -28,7 +32,20 @@ class HelpPanel extends React.Component<Props, {}> {
                 customWidth="400px"
                 closeButtonAriaLabel="Close"
             >
-                <span>{ToolTip.GetTip(this.props.tipType)}</span>
+                <div>{ToolTip.GetTip(this.props.tipType)}</div>
+                <div className="cl-help-panel-footer">
```
LarsLiden: Workable.  The solution we went with was to use Office Fabrics Panel "onRenderFooterContent" which handles this specific use case.   See:  https://github.com/Microsoft/ConversationLearner-UI/blob/master/src/components/HelpPanel.tsx

PR #866, opened 2018-12-10
feat: add scroll bar to entity picker results list (Problem #4)

mattmazzola:
Increase `maxDisplayedOptions` so more than 4 items can be shown.

Also I only chose `7em` size so that when there are more items and it has scroll bar you will see slight clipping of text letter the user know there is more below. Alternative is to make it be even cutoff 

Some thoughts about this change unrelated to the fix.  I realize it's confusing that there are only 4 entities when you actually have more, but I don't think making the list scrollable to the best solution.  I had imagined people are going to have apps with 100's of entities and didn't want them scrolling all of these in this tiny menu.  I intended people to tab into the search bar and begin typing to filter the results so that they find there desired entity filters its way to the top (within the 4).

I'm not sure what is better but an alternative is to add text saying "showing 4 out of 29 entities". Also realized we could do better to make the "New Entity" look more like a button and add the OfficeFabric + (plus) sign in front to be consistent

### DIFF
--- a/src/components/ExtractorResponseEditor/EntityPicker.css
+++ b/src/components/ExtractorResponseEditor/EntityPicker.css
@@ -53,6 +53,8 @@
     padding: 0;
     padding-bottom: 0.25em;
     border-bottom: 1px solid lightgrey;
+    overflow-y: auto;
+    max-height: 7em;
 }
 
 .custom-toolbar__result {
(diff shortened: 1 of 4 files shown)


LarsLiden:
We went with the scroll view as it proved to be the least confusing.  Wasn't obvious to users that we were only showing a subset

image

PR #867, opened 2018-12-10
 fix: prevent showing previous bot errors (Problem #6)

mattmazzola:
Clears validation errors when loading a bot since they will get reset.
Added a loading flag which blocks showing the item counts and error messages.
Kept the loading value separate than the other booleans like `invalidTraindialogs` since they have different life cycle and semantic meaning

### DIFF
--- a/src/routes/Apps/App/Index.tsx
+++ b/src/routes/Apps/App/Index.tsx
@@ -250,7 +250,7 @@ class Index extends React.Component<Props, ComponentState> {
                     <Route
                         exact={true}
                         path={match.url}
-                        render={props => <Dashboard {...props} app={app} loading={this.state.loading} validationErrors={this.state.validationErrors} />}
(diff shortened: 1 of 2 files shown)


src/routes/Apps/App/Index.tsx
```diff
@@ -33,26 +33,38 @@ import './Index.css'
...
 
 interface ComponentState {
+    loading: boolean
     validationErrors: string[];
     packageId: string;
 }
 
 class Index extends React.Component<Props, ComponentState> {
     state: ComponentState = {
+        loading: false,
```
LarsLiden: Had more or less the same solution, but used a "modelLoaded" with initialization state of "false" rather then "loading" flag.  This avoids the double set state call.  At class is recreated on navigation, one doesn't have to reset every time

PR #868, opened 2018-12-10
fix: close entity picker after creating new entity (Problem #7)

mattmazzola:
I can't remember why I didn't do this before. Hoping it was just oversight and not intentionally not adding this because it caused a problem.

### DIFF
--- a/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
+++ b/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
@@ -100,6 +100,11 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
                     }
 
                     this.props.onChangeCustomEntities([...nextProps.customEntities, newCustomEntity])
+
+                    // EntityPicker stays open after creation of new entity. Force it to close.
+                    this.setState({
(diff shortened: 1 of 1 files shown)


src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
```diff
@@ -100,6 +100,11 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
                     }
 
                     this.props.onChangeCustomEntities([...nextProps.customEntities, newCustomEntity])
+
+                    // EntityPicker stays open after creation of new entity. Force it to close.
+                    this.setState({
+                        isMenuVisible: false
+                    })
```
LarsLiden: Feels like a cleaner way to do this is closing it when the entity creator opens.  Doing it in the 
Replace this:
` onClickNewEntity={this.props.onClickNewEntity}`

This this:
`onClickNewEntity={this.onClickNewEntity}`

And add:
 ```
 @OF.autobind
    onClickNewEntity(entityTypeFilter: string) {
        this.setState({isMenuVisible: false})
        this.props.onClickNewEntity(entityTypeFilter)
    }
```

PR #869, opened 2018-12-10
fix: logic entities not properly handled

LarsLiden:
steps inserted at wrong position on wait actions


LarsLiden:
appveyor hang, restart build

PR #873, opened 2018-12-12
 feat: add warning about inconsistent entities (Problem #8)

mattmazzola:
Adds `InconsistentEntityLabellingModal`

The modal stays open until the user has 'Accepted' all the changed extract responses, Upon closing the existing `extractResponses` that are inconsistent will be replaced with correct ones that have proper labelling.

I was a little uncertain about the instructions.  Maybe you can reply and clear up if this behaves differently than expected. It didn't quite say what should trigger the modal and we don't want it to open automatically as that could be really frustrating for users. There is a UX guideline that users should always feel in control of the application and doing this automatically goes against that. I saw we had another modal/warning popping up under a condition when the user clicked "Score Actions" so I did this anyways but it's a bit misleading as we are not doing what the button says.  I thought we would be better disabling the score actions button and adding a "review labels" or "correct labels" button but those didn't look good in the space.  I ended up compromising and showing a warning similar to our text variation validation warning as seen below:
image
This at least let's the user know something is wrong and they could expect different behavior but I think it definitely needs more work because it doesn't tell them how to fix it.

I was also unsure how there could be multiple inconsistencies if the point of this is to prevent users from continuing if there is inconsistency.  I thought the only way is if they have multiple alternate extract responses but maybe there another.

Another thing I was confused about is how a user gets in this situation without editing the response first. I thought our system would detect it being the same input and predict the entities based on previously learned inputs so there would be no way to produce an inconsistency automatically.

Because of this I think an alternate solution would be to prevent the error and opt-in to changing instead of trying to correct the error after it happens. Practically this means not allowing users to edit inputs which they have already typed because it would produce an inconsistency. I imagine if a user accidentally mis-labelled an input in dialog and then reenters this input in a later dialog but realize this mistake and wants to correct it to proper labels we need to allow this without blocking. This would let the next response through and invalidate previous train dialogs instead if the new train dialog being invalid because of inconsistency.


mattmazzola:
I realized I forgot to exclude invalid train dialogs and account for different casing of input

I also went ahead with the changes I was considering above. I thing it's easier to see and use than to try to explain all the differences.

Changes:
- **Wording of warning message. "Entity labelled differently in another utterance."**
 I didn't like this message because it only explained what the state is but didn't explain what was the problem.  The user might have mis-labelled the previous input and then intended to continue with them labelled differently. I'm still uncertain about how to let them continue safely. I also wanted to change the word 'utterance'. To me 'utterance' seems like a more formal term from machine learning / vocal systems. I think we should be consistent and change all occurrences of utterance to user input.  I think this makes the product less confusing. I'm proposing "Entities must be labelled consistently in similar user inputs across all train dialogs."  This tells them what our system requires and leaves room for any type of input. (Voice or text)
- **Uses "Review Entity Labels" and disables "Score Actions"**
This avoids breaking users expectations with score actions button and explains what will happen when they click the button.
- **Now shows both extract responses so the user can compare and see mistake.**
This helps them understand the system and know what was inconsistent so they hopefully don't make the same mistake. Many pairs of names could be used such as:  "before" / "after", "existing" / "new", or "correct" / "incorrect"
A better implementation would be to find the entities that are mislabeled and color them red or something.  Imagine if there are 4 entities labeled and they are all correct but 1. If we only had the correct version it would be hard to tell what was wrong because it would look correct. 
- **Allows partial completion**
The user may accept the first 2 inconsistencies and then click close. We should still replace the responses that were accepted without closing and re-opening the modal every time they click accept.\
-- **Isolate logic bit more and Avoids changing `onClickSubmitExtractions`**
No longer needed to pass `inconsistentResponses` to it because we don't open the modal from the "Score Actions" button.

mattmazzola:
Short video of feature: http://recordit.co/hY2Cpy5ofP

src/components/modals/EntityExtractor.tsx
```diff
@@ -303,16 +306,111 @@ class EntityExtractor extends React.Component<Props, ComponentState> {
...
+        return Array.from(map.values())
+    }
+
+    /**
+     * Get all text variations that have same text but different entities
+     */
+    static getInconsistentResponses(trainDialogs: TrainDialog[], pendingExtractResponses: ExtractResponse[]): InconsistentExtractResponse[] {
+        // We assume all text variations within existing valid train dialogs are consistent
+        const textVarations = trainDialogs
+            .filter(td => !td.invalid)
```
LarsLiden: Check only happens at edit time, so seems like excluding invalid TDs could lead to inconsistent labeling.  If invalidated TD is fixed, no check will occur on the label to find the inconsistency

LarsLiden:
We opted to do the conflict detection on the server.   If there are multiple users editing train dialogs the client may have outdated information and unknowingly create conflicts.    The call to the scorer now returns a 409 - conflict with a list of the expected labels which the user can then accept or go back to editing.

PR #875, opened 2018-12-15
BROKEN feat: select words to label in train dialog (Problem #5)

mattmazzola:
I wasn't able to complete this one.  I tried the following:

1. **Using the `cy.type()` command to implicitly submit events.**
Hoped to emulate key presses that the user could have typed such as Cntrl + Shift + Right arrow them figure out the number of right arrows by the word offset and length.  Internally cypress handles all the keyup down and mouse stuff. However, it didn't seem to select the word. The first right arrow seemed to work but the entity menu closed immediately and I wasn't able to select the New Entity or existing entity button

2. **Use the `cy.trigger()` command to explicitly submit events.**
I looked up techniques used to simulate drag'n drop behavior thinking thats similar to highlighting text with your mouse. They seemed to rely and invoking the `mousedown` `mousemove` and `mouseup` events directly the correct elements.  This didn't seem to have any affect. No selection or bringing up the entity picker.

    Note: I saw an existing method in the `traindialogmodal.js` page called `highlightWord` and tried to use it which would have been a similar technique, but it didn't work eitherl

3. **Use the `cy.then()` to get direct access to `.slate-editor` element**
Since the above two approaches didn't work I was resorting to an even more low-level API. I thought I could manually build and set the browser selection / range.  It seems to select the entire node contents using `selectNodeContents` and didnt work when I tried to use the more specialized `setStartBefore` and `setEndAfter` to select only the specified words.

  Even tried just selection `.cl-token-node` containing the word "hovercraft" but that didn't work either.

### DIFF
--- a/cypress/integration/entityLabel.ts
+++ b/cypress/integration/entityLabel.ts
@@ -1,0 +1,65 @@
+/**
+* Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+*/
+const { convLearnerPage,
+  modelpage,
+  entity,
+  entityModal,
+  actions,
+  actionsModal,
+  trainDialogPage,
+  trainDialogModal,
+  scorerModal,
+  logDialogPage,
+  logDialogModal,
+  testLog } = components();
+
+/**
+* Wait action: After the system takes a "wait" action, it will stop taking actions and wait for user input.
+* Non-wait action: After the system takes a "non-wait" action, it will immediately choose another action (without waiting for user input)
+*/
+describe('Label text in train dialog', function () {
+  const postfix = Cypress.moment().format("MMMDD-HHmm")
+  const modelName = `e2e-entity-${postfix}`
+
+  it('create a New Model', function () {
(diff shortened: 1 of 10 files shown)


LarsLiden:
Yeah, this was a hard one.  Even reached out to cypress support and they said that they didn't have a way to solve it.  In the end, I was able to create a custom event that we can call to select the text.  See [ExtractorResponseEditor.tsx](https://github.com/Microsoft/ConversationLearner-UI/blob/b46002ddfc35292376e4bd65e9267efb000b97e6/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx) for the solution.

For example, to select the words "the best", Michael does the following:
```
var event = new CustomEvent("Test_SelectWord", {detail: "the best"})
document.dispatchEvent(event)
```

cypress/integration/entityLabel.ts
```diff
@@ -0,0 +1,65 @@
+/**
```
LarsLiden: Nice that you did the tests too.

PR #883, opened 2018-12-20
fix: bug 1734 - popup overlaps delete icons

pnwbrian:
Reworked the styling if the popup is due to overlapping, duplicate entities


LarsLiden:
I eliminated this pop-up so this isn't needed any more

PR #884, opened 2018-12-21
fix: bug 1758 - block second "Master" version tag

pnwbrian:
Block any further attempts to create another "master"


src/components/modals/PackageCreator.tsx
```diff
@@ -128,20 +119,20 @@ class PackageCreator extends React.Component<Props, ComponentState> {
...
-                                    id: FM.PACKAGECREATOR_LIVE_LABEL,
-                                    defaultMessage: 'Make Live Version'
-                                })}
-                                checked={this.state.isLiveVal}
-                                onChange={this.onToggleSetLive}
-                                tipType={ToolTips.TipType.PACKAGECREATOR_LIVE_TOGGLE}
-                            />
-                        </div>
+                        <TC.Checkbox
+                            label={intl.formatMessage({
```
LarsLiden: Could use formatMessageId in util.ts

PR #885, opened 2018-12-21
Feature 1698 - disable Expected Entities field for non-wait actions

pnwbrian:
Request mentioned hiding, but we collectively agree disabling better than hiding


src/components/ExtractorResponseEditor/EntityPicker.tsx
```diff
@@ -30,7 +30,8 @@ export default class EntityPicker extends React.Component<MenuProps> {
         const style: any = {
             left: this.props.isVisible ? `1em` : null,
             bottom: this.props.isVisible ? `${this.props.position.bottom}px` : null,
-            height: "14em"
+            height: !this.props.isOverlappingOtherEntities ? "14em" : "4em",
```
LarsLiden: Looks like another PR got mixed in with this one

PR #893, opened 2018-12-26
fix: bug 1776 invalid warning about unconnected entities

LarsLiden:


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -374,8 +374,9 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                 key: k,
                 current: v,
                 prev: prevValue,
-                currentEntities: ActionPayloadEditor.Utilities.getEntitiesFromValue(v),
-                prevEntities: ActionPayloadEditor.Utilities.getEntitiesFromValue(prevValue)
+                // TODO: Should these be be getAllEntitiesFromValue
```
LarsLiden: @mattmazzola  - this seems like it should be over all entities, not just the non-optional ones, but wanted you to confirm before changing?

PR #895, opened 2018-12-26
fix: bug 1758 - block second "master" tag

pnwbrian:
Reworked the name checking logic a bit, and add name checking to ENTER/keydown logic

### DIFF
--- a/src/components/modals/PackageCreator.tsx
+++ b/src/components/modals/PackageCreator.tsx
@@ -48,14 +48,14 @@ class PackageCreator extends React.Component<Props, ComponentState> {
 
     @OF.autobind
     onClickCreate() {
-        this.props.onSubmit(this.state.tagNameVal, this.state.isLiveVal)
+        this.props.onSubmit(this.state.tagNameVal.trim(), this.state.isLiveVal)
     }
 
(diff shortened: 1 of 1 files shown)


public/manifest.json
```diff
@@ -4,12 +4,12 @@
   "icons": [
     {
       "src": "favicon.ico",
-      "sizes": "192x192",
```
LarsLiden: Looks like maybe you are forgetting up update your local branch before starting on other work or working off the wrong branch?  Means that work items are getting mixed.  This one was already merged back into master...

src/components/modals/PackageCreator.tsx
```diff
@@ -55,7 +55,7 @@ class PackageCreator extends React.Component<Props, ComponentState> {
     // Also has benefit of native browser validation for required fields
     onKeyDown(event: React.KeyboardEvent<HTMLElement>) {
         // On enter attempt to create the model if required fields are set
-        if (event.keyCode === 13 && this.state.tagNameVal) {
+        if (event.keyCode === 13 && (!(this.onGetNameErrorMessage(this.state.tagNameVal)).length)) {
             this.onClickCreate();
```
LarsLiden: I'd trim in onClickCreate too so that the created model has no spaces

PR #900, opened 2018-12-29
Remove Markup Bug Fix

SDESkowronski:
Chat messages recently had HTML added to them, I had to remove the markup in order to validate the text.

### DIFF
--- a/cypress/support/Helpers.js
+++ b/cypress/support/Helpers.js
@@ -25,12 +25,25 @@ export function RemoveDuplicates(inputArray)
   return uniqueOutputArray
 }
 
-export function StringArrayFromInnerHtml(selector) 
+export function RemoveMarkup(stringWithHtml)
+{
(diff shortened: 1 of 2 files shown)


SDESkowronski:
While I agree with your suggestion for product code, I do not agree with it for test code because the intended audience of the resulting application is different.

For test code projects I have been on, I rarely see checks on parameters. In the case of this code if you don’t pass in the selector then the code will not work, putting in an extra check will not really add any value, except to change the error message, but the reason for the failure should be obvious to the developer in either case.

Putting tests on all parameters for all functions also adds extra time to the test as well as complicates the code. Test code always needs to be as fast as it can be.

LarsLiden:
I think I agree with Michael here.  The test code is deterministic (meaning the same code path is followed every time) so an invalid parameter will always cause the test to fail.

PR #904, opened 2019-01-02
feat: warning and reload bot info if running bot changed

LarsLiden:


LarsLiden:
re-trigger travis build

PR #910, opened 2019-01-03
fix: fea 1699 - enhance model deletion dialog confirmation

pnwbrian:
Added richer HTML dialog mesage support and button styling support. Reworked the text a bit as well


src/react-intl-messages.ts
```diff
@@ -1025,6 +1028,9 @@ export default {
         [AT.EDIT_TRAINDIALOG_ASYNC]: 'Editing Train Dialog',
         [AT.EDIT_APP_LIVE_TAG_ASYNC]: 'Editing Application live tag',
         [AT.EDIT_APP_EDITING_TAG_ASYNC]: 'Editing Application editing tag',
+
+        // General UX strings not fitting elsewhere
+        [FM.APP_MESSAGE_PERMDELETE]: 'Deletion is irreversible.',
```
LarsLiden: Still would have called this "FM.APPSLIST_MESSAGE_PERMDELETE" and put it with the other one.   (These are in the file alphabetically too)

pnwbrian: Folded things into inline JSX to support bold model name

pnwbrian:
Moving the delete model action under settings now

PR #915, opened 2019-01-07
Validate error reporting

SDESkowronski:
1st test case to validate the error triangles and messages that occur during training.


cypress/tests/EditAndBranching.js
```diff
@@ -91,3 +89,34 @@ export function TagAndFrog()
...
+{
+  models.ImportModel('z-errorHandling', 'z-disqualifyngEnt.Trained.cl')
+  modelPage.NavigateToTrainDialogs()
+  cy.WaitForTrainingStatusCompleted()
+  
+  modelPage.VerifyNoErrorIconOnPage()
+
+  train.EditTraining('Hey', 'world peace', "Sorry $name, I can't help you get $want")
+  editDialogModal.InsertUserInputAfter('Sam', 'InsertedText')
+  editDialogModal.VerifyErrorMessage('This Train Dialog has errors that must be fixed before it can be used to train your model')
```
LarsLiden: Wonder if at some point we should use references to the string in our strings table so if we change them it doesn't break the tests

SDESkowronski: Yes, I will do that after I get done with this DeleteAllModels test case.

PR #917, opened 2019-01-10
feat: task 1834 - extend app list with model id

pnwbrian:
Extending app list with model id for improved test automation


SDESkowronski:
Works for me. Thanks.

PR #919, opened 2019-01-11
Delete All Test Models Update

SDESkowronski:
Fixed the Delete All Test Models so that it uses the API to do this job.


cypress/integration/tools/zTemp.js
```diff
@@ -21,4 +21,5 @@ describe('zTemp test', () =>
     // cy.pause()
     // cy.Train_CaptureAllChatMessages()
   })
-})
\ No newline at end of file
+})
```
LarsLiden: As a reminder this won't get pushed to NPM unless you have the prefix "fix" or "feat" in the check-in name.  Pull request accepted but no new package created.  You'll need to create another PR to update the package

PR #920, opened 2019-01-11
fix: 1831 insert and button issues

LarsLiden:


src/components/modals/EditDialogModal.tsx
```diff
@@ -241,21 +241,25 @@ class EditDialogModal extends React.Component<Props, ComponentState> {
...
+            // Can insert here but would be inconsistent with TeachSession behavior
+            /* 
+            const buttonSubmit = activity.channelData && activity.channelData.imback
+            if (this.state.selectedActivity && buttonSubmit) {
                 await this.props.onInsertInput(this.state.currentTrainDialog!, this.state.selectedActivity, userInput.text, this.state.addUserInputSelectionType)
             }
+            */
             // If next action should be score, need to insert, not continue
-            else if (this.waitingForScore()) {
+            if (this.waitingForScore()) {
```
LarsLiden: This feels super hacky, but webchat doesn't differentiate between button clicks and clicking on an action when posting a select activity call, so I have to track by noting that the previous input was an imback

PR #922, opened 2019-01-11
fix: bug 1835 - boldify end session actions in listviews

pnwbrian:
Extended className property to support additional CSS class selectors, including a new "cl-ux-bolditalic" for all session action listviews

### DIFF
--- a/src/components/actionPayloadRenderers/SessionPayloadRendererContainer.tsx
+++ b/src/components/actionPayloadRenderers/SessionPayloadRendererContainer.tsx
@@ -26,6 +26,7 @@ export default class Component extends React.Component<Props, {}> {
 
         return (
             <TextPayloadRenderer
+                classNameEx="cl-ux-bolditalic"
                 original={renderStringUsingEntityNames}
                 currentMemory={renderStringUsingCurrentMemory}
             />)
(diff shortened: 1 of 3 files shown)


src/components/actionPayloadRenderers/SessionPayloadRendererContainer.tsx
```diff
@@ -26,6 +26,7 @@ export default class Component extends React.Component<Props, {}> {
 
         return (
             <TextPayloadRenderer
+                classNameEx="cl-ux-bolditalic"
                 original={renderStringUsingEntityNames}
```
LarsLiden: Rather than pollute the TextPayloadRenderer, why not create a "SessionPayloadRenderer" akin to the "ApiPayloadRenderer" and "CardPayloadRenderer".  This give us more flexibility as well as make it easy to support the multi-line thing

pnwbrian: I considered this idea, then figured we might want to enhance more text rendering down the road. If true, extending the props on the text rendering opens styling up to everyone, not just session rendering

LarsLiden:
Yes, but in this case we want more than just styling – i.e. multi-line

pnwbrian:
Superseded by later PR/code changes

PR #923, opened 2019-01-11
fix: bug 1819 - icon hover state restyled to bi-chromatic, from tri

pnwbrian:
Swapped out the .ms-Button-icon:not(:disabled):hover styling from a dark (blackish) color, to an orange-ish color that looks better on white and the red background of errors and warnings. And sync'd colors with the cl-ux-redbackground hover CL custom class

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -132,7 +132,7 @@ html {
 }
 
 .cl-icon-redbackground:hover {
-  color:orange !important;
+  color:rgb(252, 167, 9) !important;
   background-color:transparent !important;
 }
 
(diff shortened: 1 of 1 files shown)


public/index.css
```diff
@@ -209,7 +209,7 @@ html {
 }
 
 .ms-Button-icon:not(:disabled):hover {
-  color: var(--color-themeDark) !important;
+  color:rgb(252, 167, 9) !important;
```
LarsLiden: Please add this as a color to the color table rather than refer to RGB values directly.  Thanks!

PR #925, opened 2019-01-12
fix: bug 1842 - web chat active selection background

pnwbrian:
Switched to orange/maroon palette for active selection

- Queued for team review


mattmazzola:
I find it easier to review things like this if there is a visual. I know we move fast and it's not always necessary to take the pictures but here there for anyone else who needed to see the change:

**Before:**
image

**After:**
image

I think it is an improvement because the red made it look like it was highlighted due to error.

Unrelated to this specific change but my thoughts about web chat look:

- Prefer to get rid of the stroke around the User Bot buttons. It makes the buttons look blurry or unfocused. I see that the strke is also the color of the "+" sign inside so this makes it difficult.

Example:
image

- Would maybe put white circle or something behind the trashcan to have more clickable boundaries of a button and be more visible. Red loses contrast on the orange. 

Example:
image

pnwbrian:
Deferring trash icon changes

pnwbrian:
Deferring lower priority trash can icon changes

PR #927, opened 2019-01-15
New Test Case - UserInputFollowsNonWaitActionErrorHandling

SDESkowronski:


cypress/support/components/EditDialogModal.js
```diff
@@ -205,3 +223,20 @@ export function InsertUserInputAfter(existingMessage, newMessage)
   cy.RunAndExpectDomChange(() => { Cypress.$('[data-testid="chat-edit-add-user-input-button"]')[0].click() })
   cy.Get('[data-testid="user-input-modal-new-message-input"]').type(`${newMessage}{enter}`)
 }
+
+export function InsertBotResponseAfter(existingMessage, newMessage)
+{
+  SelectChatTurn(existingMessage).then(indexOfSelectedChatTurn =>
+  {
+    var indexOfInsertedBotResponse = indexOfSelectedChatTurn + 1
+    cy.RunAndExpectDomChange(() => { Cypress.$('[data-testid="chat-edit-add-bot-response-button"]')[0].click() })
```
pnwbrian: I'm thinking the [0] index of the "chat-edit-add-bot-response-button" selection would insert this new message at the first instance of the matching existing message. Normally this should be fine, but I've been spoiled by picking the "n" occurrence of the same message. Just a suggestion

SDESkowronski: I believe you misunderstood, that is if I understand your comment correctly...

That is a jQuery call, it ALWAYS returns an array of elements, but in this case it will return only 1 element, so on that line of code 'n' is ALWAYS zero.

HOWEVER...I will add an index as a parameter to the function this code is in, it needs to pass the index to the SelectChatTurn function call. This fix will be in the next iteration of this test case, I've been expanding it rather than writing a nearly duplicate test case with a slight change.

PR #933, opened 2019-01-15
feat: feature 1835: align session payload renderer with text,api

pnwbrian:
Visually align the presentation of session actions to match api and text payload rendering. Cleaned up some stale css class names in the text renderer while working in the area


src/components/actionPayloadRenderers/SessionPayloadRenderer.tsx
```diff
@@ -25,11 +25,14 @@ export default class Component extends React.Component<Props, {}> {
             ? null
             : sessionAction.renderValue(Util.createEntityMapFromMemories(entities, memories), { fallbackToOriginal: true })
 
-        return <div className={`${OF.FontClassNames.mediumPlus} cl-text-payload cl-font--emphasis`}>
-            <div className="cl-text-payload__string" data-testid="action-scorer-text-response">
-                EndSession: {(renderStringUsingCurrentMemory === null)
-                    ? `${renderStringUsingEntityNames}`
-                    : `${renderStringUsingCurrentMemory}`
+        return <div className={`${OF.FontClassNames.mediumPlus}`}>
+            <div className="cl-payloadrenderer-session-primary" data-testid="action-scorer-session-response">
```
mattmazzola: Both of these `data-testid` attributes are `action-scorer-session-response` they should be unique through application to guarantee tests select the right item.

If could put the `data-testid` attribute on the parent `<div>` to only have one. Or maybe separate into system and data responses

pnwbrian: Good catch. I've appended "-extended" to the second, optional DIV as that's extended session data.

mattmazzola: In my opinion the word `extended` is too generic. It doesn't explain what the extension is or why it's extended. None our other renderers have this concept of extended either so there is no understood meaning. I would prefer something like `-system` and `-user` since `EndSession` is hard coded and won't change and the other field is optional and from the user.

That all being said it's not that important. Only a test id, doesn't effect product function, and isn't visible to customer so I'd still approve.

pnwbrian: Good suggestions. I'm going to switch to -user as that capture's the intent of the code block

PR #934, opened 2019-01-15
fix: relax Poller test constraints to increase reliability

mattmazzola:
I think the issue might be related the differences in processing speed CI environments or variability in SetTimeout calls. Different environments can adjust them by a couple milliseconds so they're only approximate.
The odd thing is in this case it was *faster* instead of *slower* which you would not expect since they could throttle NodeJS timers or something.
The poller starts and has fixed interval of 100, we then add a poll immediately with expiration of 500.
The poller should invoked the methods at 100, 200, 300, 400
At 500 it should be expired and not call those methods thus leaving the count at 4, but somehow it was called 1 more time.
This could only happen if the last poll time was something like ~499

### DIFF
--- a/src/services/poller.test.ts
+++ b/src/services/poller.test.ts
@@ -49,9 +49,9 @@ describe('Poller', () => {
         const poller1 = new poller.Poller({ interval: 100 })
         await poller1.addPoll(pollConfig)
 
-        expect(requestMock.mock.calls.length).toBe(4)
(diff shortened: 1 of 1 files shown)


mattmazzola:
The `npm test` which runs the `poller.test.ts` succeeded, but it failed due to unrelated end-to-end / Cypress tests.

PR #937, opened 2019-01-16
fix: TEXT: bug 1853 - improve tool tip grammar and spelling

pnwbrian:
Text change for clarity and spelling typo


src/components/ToolTips/ToolTips.tsx
```diff
@@ -326,8 +326,9 @@ export function getTip(tipType: string) {
             return (
                 <div>
                     <h2>Replay is needed</h2>
-                    <p>One or more edits to Action or Entity properties, may have invalidated turns in this conversation</p>
-                    <p>It must be replayed to see if any errors were introduced before it can be used in trianing</p>
+                    <p>One or more turns in this conversation may have been invalidated by edits
+                        to Action or Entities. Replay this conversation so it may be used in
+                        training as errors were introduced.</p>
```
mattmazzola: This second sentence seems off to me. It almost implied errors WERE introduced but I thought we we're only replaying to find out if errors were introduced. I think only thing we know is there was a change to bot code, but we don't know if it's compatible or not.

Minor, but I always preferred the words "input" and "response" over "turns".  Probably doesn't work in this help text because it would be too wordy but just a thought.

PR #944, opened 2019-01-17
Automation end session actions

pnwbrian:
End session rendering was tripping up test automation as the element's markup is quite different than text actions. Now the code scans the element DOM looking for adaptive card type markup, and skips for now. Standard text rendering verification/asserts remain


cypress/support/components/ScorerModal.js
```diff
@@ -18,14 +18,22 @@ export function ClickAction(expectedResponse)
...
-    cy.wrap(elements[elements.length - 1]).within(e => {
-      cy.get('div.format-markdown > p').should('have.text', expectedUtterance)
-    })})
+  cy.Get('[data-testid="web-chat-utterances"]').then(elements => 
+  {
+    var element = elements[elements.length - 1]
+    if (!Cypress.$(element).find('div.wc-adaptive-card').length) {
+      expect(helpers.RemoveMarkup(Cypress.$(element).find('p').text())).to.equal(expectedUtterance)
+    } else {
+      if (expectedUtterance && (Cypress.$(element).html().indexOf(helpers.RemoveMarkup(expectedUtterance)) < 0)) {
```
SDESkowronski: This is backwards...You need to remove the markup from the UI Element NOT from the "expectedUtterance" that was basically passed into this method.

SDESkowronski: Actually lines 29, 31 & 32 can go away and it should work for all cases.

pnwbrian:
deleting to clean up code formatting

PR #945, opened 2019-01-17
chore: increase type coverage for cypress

mattmazzola:
Ideally we would have full type coverage because everything would be Typescript but that is more complicated and I'm still investigating. This is preliminary change that still has a lot of value.

Before there were NO types.
See below I'm hovering over `cy` and it says `any` which is not useful
image

**1. Add `/// <reference types="Cypress" />` to one of the files.**
I originally thought it had to be added to all files but it seems only one is ok. I only put it one one to minimize changes but we can add it future if needed.

I'm hovering over `cy` but get all the type information
image

This is great, but as you can see even with the `/// <reference>` the incorrectly cased method names are not found.
image

**2. Fix the method name casing**
image

## Outcome:
- Have type information and standard `cy` chainable method.

## Does NOT cover:
- Commands
There is extensive use of custom commands like `WaitForStableDOM` which is not known about.
After this, I believe everything will be `any`

## Future plans:
- Convert everything to Typescript

## Going forward:
Please use the `camelCased` names the are available with typing information.  I'm not even sure how the Pascal case names worked before. I assume cypress has an alias. Anyways, it is a convention for Typescript / Javascript.  Capital method names are usually reserved for `static` methods.

### DIFF
--- a/cypress/support/MonitorDocumentChanges.js
+++ b/cypress/support/MonitorDocumentChanges.js
@@ -8,7 +8,7 @@
  * cy.route() commands used only for waiting. It does this by constantly monitoring the
  * DOM for changes and tracking the Milliseconds Since the Last Change and also resetting
  * those Milliseconds to zero when we see the spinner is displayed. Certain cy.commands(),
- * such as cy.Get(), are then modified to prevent execution until this Millisecond count
+ * such as cy.get(), are then modified to prevent execution until this Millisecond count
  * reaches at least 700.
  * 
  * The basic premis this works on is that when the application under test is making API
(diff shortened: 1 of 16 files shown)


mattmazzola:
Was there anything special about `cy.Get` vs `cy.get` ? 

I did not see any custom commands named `Get` and thought it was only a casing issue but perhaps it was doing more? @SDESkowronski

mattmazzola:
Talked offline. Saw those ARE infact custom commands. Must have searched wrong term and incorrectly concluded it was only casing. Closing issue for now as it's won't really help at all.

Since `cy.Get` is a custom command and everythin after will likely be `any` and `cy.Get` is usually get the first of any test step it will still be `any` even with the changes

PR #947, opened 2019-01-21
Automation end session actions

pnwbrian:
Added the Session action click handler and adaptive card chat message verification bit so the ENDSESSION test is complete and passes


cypress/support/Train.js
```diff
@@ -126,6 +126,16 @@ export function SelectAction(expectedResponse, lastResponse)
   })
 }
 
+export function SelectSessionAction(expectedResponse, lastResponse)
```
SDESkowronski: I would prefer that the name of this method AND the data_testid BOTH use End in the name, ie…"SelectEndSessionAction" and "action-scorer-end-session-response". The goal is for variable and function names to clearly match the UI elements they represent and manipulate.

cypress/support/components/ScorerModal.js
```diff
@@ -6,45 +6,59 @@
...
-export function ClickAction(expectedResponse, expectedIndexForActionPlacement)
-{
+export function ClickAction(expectedResponse, expectedIndexForActionPlacement) {
   cy.Get('[data-testid="action-scorer-text-response"]').ExactMatch(expectedResponse)
     .parents('div.ms-DetailsRow-fields').find('[data-testid="action-scorer-button-clickable"]')
     .Click()
   VerifyChatMessage(expectedResponse, expectedIndexForActionPlacement)
 }
 
+export function ClickSessionAction(expectedResponse, expectedIndexForActionPlacement) {
```
SDESkowronski: Click**End**SessionAction please

cypress/support/components/ScorerModal.js
```diff
@@ -6,45 +6,59 @@
...
-    })})
+    })
+  })
+}
+
+export function VerifyAdaptiveChatMessage(expectedMessage, expectedIndexOfMessage) {
+  var expectedUtterance = expectedMessage.replace(/'/g, "’")
+  cy.Get('[data-testid="web-chat-utterances"]').then(elements => {
+    if (!expectedIndexOfMessage) expectedIndexOfMessage = elements.length - 1
+    cy.wrap(elements[expectedIndexOfMessage]).within(e => {
```
SDESkowronski: I'd prefer "element" here instead of "e". I would really like to retain the verbose naming, it is more readable.

SDESkowronski: I don't see how this actually tests that the expected response appears in the chat window, the expectedMessage is never used.

cypress/support/Train.js
pnwbrian: While agreed, I wrote the code to supports any adaptive card - e.g. end session, api and form type actions so I opted for the broader adaptive card type chats. The hard coded "end session" code disappears, to be passed as the arg so double win

cypress/support/components/ScorerModal.js
pnwbrian: Good catch. I opted to keep the test generic know in favor of getting something in the tree as the end session issue was a showstopper. I'll double back in the Verify code once my API and form code tests are ready this week

PR #948, opened 2019-01-22
feat: allow interface to be served at /ui path

mattmazzola:
Currently we require a separate command to serve the UI. This runs the bot on a separate port from the bot and serves it at the root.

Goals:
- This PR will change the package output so the UI can be served at `/ui` path.
- Development remains unchanged! You can still `npm start` and view at `localhost:5050/`
- Tests run from CircleCI should remain unchanged because they use 

Also added a `server` folder which allows testing the built package and seeing that it works on the `/ui` path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ package
 .env.test.local
 .env.production.local
 ~*.xlsx
+server.js
 
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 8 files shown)


LarsLiden:
Potentially doing one 1st week in Feb to get new changes out.  In general think it would be good to update docs in || with code changes so we don’t forget or have a big backlog just before release

mattmazzola:
Oh ok.

Yep, certainly agree all changes for a feature should be in a single PR.

PR #952, opened 2019-01-24
fix: new UX test block

pnwbrian:
New test block for generic UX/UI testing code


cypress/tests/UX.js
```diff
@@ -0,0 +1,12 @@
...
+* Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+*/
+
+const models = require('../support/Models')
+const helpers = require('../support/Helpers')
+
+export function BotModelMismatch() {
+  models.ImportModel('BotModelMismatch', 'z-botmodelmismatch.cl')
+  assert(Cypress.$(".cl-errorpanel").length < 1, "Bot/Model mismatch should display error msg!")
```
SDESkowronski: This code (line 11) belongs in cypress\support\components\ModelPage.js. Notice the pattern, we don't put any code in the test files that refer to UI elements, nor do we put complex logic in these test files either. The test files should be easy to read for anyone wanting to understand the test scenario, thus we only call well named functions here that describe what the test step is doing.

PR #953, opened 2019-01-24
Parallel Test Runs

SDESkowronski:
Changes to support running our test cases in parallel.

NOTE: I did not merge the current test code in. This is because it underwent a radical reformatting and I don't want the noise in this code review. Once these changes have been approved I'll merge in the latest code from Brian.


cypress/plugins/index.js
```diff
@@ -16,19 +16,34 @@ module.exports = (on, config) => {
...
-  //     args.push('--disable-blink-features=RootLayerScrolling')
-  //     return args
-  //   }
-  // })
+  on('before:browser:launch', (browser = {}, args) => {
+    //require('../GenerateCypressTestSpecs')
+    
+    // BUG BUG - Need to work out why this code does not execute or where it is throwing an exception.
+    alert('got here');
+    var fs = require('../../node_modules/pn/fs')
```
mattmazzola: This path looks strange. Why did you use this one? 

Normally for global modules you only specify the module name like:
`var fs = require('fs')` OR better `import * as fs from 'fs'`

cypress/plugins/index.js
```diff
@@ -16,19 +16,34 @@ module.exports = (on, config) => {
...
       console.log(message)
       return null
     }, 
     parse: (filePath) => {
       return path.parse(path.normalize(filePath))
-    }
+    },
+    exists: (path) => {
+      return new Promise(resolve => { 
+        fs.access(path, fs.constants.F_OK, (err) => { resolve(err ? false : true) })
```
mattmazzola: It's weird to use `fs` variable when it's not declared in scope. The seems to only leverage `var` being global and assigned by the time this is called.  I would think it could be `undefined`

cypress/support/TestListManager.js
```diff
@@ -3,225 +3,140 @@
...
 
   // groupName.testName - 'testName' from 'groupName'
   // TODO: Add support for these wild card versions
   // *.*                - All Groups, All Tests
   // *.testName         - All tests with all groups matching 'testName'
   // groupName.*        - All tests from 'groupName'
-  get next()
-  {
-    if (this.index >= this.testList.length) return undefined
+  get next() {
```
mattmazzola: Given that there is already an iterator protocol and convention it's confusing to invest yet another.

I'd at least stick with `next()` as a regular function instead of a getter method.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next

cypress/support/TestListManager.js
```diff
@@ -3,225 +3,140 @@
...
-function GetTestGroup(name)
-{
-  for (var i = 0; i < testGroups.length; i++)
-  {
-    if (testGroups[i].name == name) 
-    {
-      return testGroups[i]
-    }
+function GetTestGroup(name) {
+  for (var i = 0; i < testGroups.length; i++) {
```
mattmazzola: I would avoid the `for` loops and instead use the safer array methods that clearly explain what your trying to do and reader can know what should happen in the loop. 

Example:
`testGroups.find(testGroup => testGroup.name === name)`

cypress/support/TestListManager.js
```diff
@@ -3,225 +3,140 @@
...
-function GetTest(testGroup, testNameToFind)
-{
-  var toFind = `function ${testNameToFind}(`
-  for (var i = 0; i < testGroup.tests.length; i++)
-  {
-    if (`${testGroup.tests[i].func}`.startsWith(toFind)) 
-    {
-      return testGroup.tests[i]
+function GetTest(testGroup, testNameToFind) {
+  for (var i = 0; i < testGroup.tests.length; i++) {
```
mattmazzola: again use find and only pass minimal info to function. It doesn't need the whole `testGroup` only the `tests`

Example:
`tests.find(test => test.func.toString().substring(9).startsWith(testName))`

cypress/tests/zTemp.js
```diff
@@ -0,0 +1,21 @@
...
+const helpers = require('../support/Helpers')
+const homePage = require('../support/components/HomePage')
+const modelPage = require('../support/components/ModelPage')
+const train = require('../support/Train')
+const trainDialogsGrid = require('../support/components/TrainDialogsGrid')
+const editDialogModal = require('../support/components/EditDialogModal')
+
+Cypress.TestCase('zTemp', 'Temporary Experimental Test', zTemp)
+function zTemp() {
+  cy.task('exists', 'C:/repo/ConversationLearner-UI/cypress/integration/Tools').then(fileExists => {console.log(`fileExists: ${fileExists}`)})
```
mattmazzola: Shouldn't use hard-coded paths. Only works on your machine

cypress/GenerateCypressTestSpecs.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+  });
+}
+
+function CreateSpecFile(groupName, testName) {
+  var folderPath = `./cypress/integration/${groupName}`;
+  if (!fs.existsSync(folderPath)) { fs.mkdirSync(folderPath) }
+  
+  var filePath = `${folderPath}/${testName}.js`;
+  if (!fs.existsSync()) {
```
mattmazzola: Looks like your forgot the argument to the function. I think you intended to write a file if it doesn't already exist. \

`!fs.existsSync(filePath)`

I'm not sure how this works. Maybe it always returns `false` if a path isn't given so it was always writing the files and you never noticed.

cypress/GenerateCypressTestSpecs.js
```diff
@@ -0,0 +1,109 @@
...
+//  Cypress.TestCase(testGroupName, testDescription, testFunction)
+//
+// It uses that data to create:
+//   1) The group subfolders within "cypress/integration".
+//   2) An individual test spec file for each, in that same folder as 
+//      expected by Cypress.
+//   3) The test list: masterListOfAllTestCases
+// ----------------------------------------------------------------------
+
+var fs = require('../node_modules/pn/fs')
```
mattmazzola: This is risky and should be avoided. It seems your digging into the `node_modules` folder and assuming `pn` will be there, but it may not.  If any of dependencies decide to use a different package we would break. We're only supposed to depend on packages we explicitly declare.

```
npm ls pn
@conversationlearner/ui@0.0.0-development D:\repos\cl-ui
`-- react-scripts-ts@3.1.0
  `-- ts-jest@22.0.1
    `-- jest-config@22.4.4
      `-- jest-environment-jsdom@22.4.3
        `-- jsdom@11.12.0
          `-- pn@1.1.0`
```

At the very least I wonder if `var fs = require('pn/fs');` works?

cypress/GenerateCypressTestSpecs.js
```diff
@@ -0,0 +1,109 @@
...
+  else console.log('NOTHING HAS CHANGED')
+}
+
+function ParseTestJsFile(fileName) {
+  var fileContents = fs.readFileSync(`${pathToTestFiles}/${fileName}`, 'ascii')
+  //console.log(`File ${fileName}:\r\n${fileContents}`)
+  
+  var fileLines = fileContents.split('\n')
+  
+  fileLines.forEach(line => {
```
mattmazzola: I don't understand why this needs to be done. This is coupling this code of parsing a file with specific function `Cypress.TestCase(...)`.  If we `Cypress.TestCase` or forget to add this line the test won't work.

This is making too many assumptions. It is magic that is hard for people to understand and is not natural. We want people to be able to add a spec file, use your helper methods, and have it work as expected.

cypress/plugins/index.js
SDESkowronski: That code still needs to be fixed, but it does not appear to be influencing the test runs in anyway so I left it in for the next round of work on it. 

However, when I did have something working in this plugin file I had to do the require that way, I am not sure why, but it was the only way that worked. There are many things that Cypress is doing that interferes with "normal" JavaScript or Node coding.

cypress/GenerateCypressTestSpecs.js
SDESkowronski: That one I just now fixed. It was an artifact of trying to get a version of this code running under Cypress.

cypress/GenerateCypressTestSpecs.js
SDESkowronski: Good catch, thanks.

cypress/tests/zTemp.js
SDESkowronski: This is a temporary test case, it is used to experiment, it was not intended for others to run it. Normally I would not use a hard-coded path like this. However I did just fix it.

cypress/plugins/index.js
SDESkowronski: I commented out that code for now as well.

SDESkowronski: Ah, I just figured out the issue with this file and the troubles I've had with it. You must completely close and restart Cypress for changes in this file to take affect. Which explains why the require gave me troubles originally. I have fixed it.

cypress/plugins/index.js
SDESkowronski: Fixed

cypress/support/TestListManager.js
SDESkowronski: Originally I had used another technique for the iterator, one I found online for this pattern, however, it did not work with Cypress for some reason and I spent many hours trying to figure out why. For this PR, I will check this code in as is. I will take another look at the JavaScript iterator pattern and try once again to make it work. At a minimum I'll convert "next" to a regular function as requested.

cypress/support/TestListManager.js
SDESkowronski: fixed

cypress/support/TestListManager.js
SDESkowronski: Fixed. Thanks for pointing these out, being new to JavaScript there are still a lot of helpful functions and patterns that I don't know about.

mattmazzola:
I'll have to submit a dummy PR to test the review system, but currently it seems weird. Either I'm using it wrong or it seems like you are commenting that things are fixed or marking things resolved before you actually push your changes with the fixes. When I open a "Show resolved" item the code looks the same.

I think you should only mark things as resolved when you've actually resolved them. Otherwise we get notifications to look at the code but nothing has changed. We see the outdated stuff even though you may have fixed it on your local machine.

I'm not used to GitHubs review system. On internal Microsoft stuff I generally use "Pending" state to let people know I've acknowledged the issue and am working on it. They should expect a change. If it's not in pending it means I still need more opinions or discussion from reviews to decide what to do.  I don't think GitHub has this though.

SDESkowronski:
>           I'll have to submit a dummy PR to test the review system, but currently it seems weird. Either I'm using it wrong or it seems like you are commenting that things are fixed or marking things resolved before you actually push your changes with the fixes. When I open a "Show resolved" item the code looks the same.
> I think you should only mark things as resolved when you've actually resolved them. Otherwise we get notifications to look at the code but nothing has changed. We see the outdated stuff even though you may have fixed it on your local machine.
> I'm not used to GitHubs review system. On internal Microsoft stuff I generally use "Pending" state to let people know I've acknowledged the issue and am working on it. They should expect a change. If it's not in pending it means I still need more opinions or discussion from reviews to decide what to do.  I don't think GitHub has this though.

I get your frustration. I had not yet pushed my changes in. I was making the changes locally and then replying to you, now that I've made the changes and tested them all I've pushed them in.

cypress/support/TestListManager.js
```diff
@@ -125,18 +125,9 @@ class TestListIterator {
 }
 
 function GetTestGroup(name) {
```
mattmazzola: This is a bit misleading. Would prefer to name the method in a way that indicates how it will behave. `findTestGroup`
It is named `Get` but internally uses `find`. These have different meanings.

To me `get` prefixed methods are like a getter. It should be guaranteed to get the result of the function. Otherwise something is wrong and it is an error. In which case it should `throw`

`find` prefixed method imply the are searching or trying to find something. It is not guaranteed they will succeed and in the case they don't find it they return `undefined`

Even the language authors chose `find` instead of `get` for this type of method.  I think we should preserve their naming to avoid confusion.

cypress/GenerateCypressTestSpecs.js
```diff
@@ -0,0 +1,121 @@
...
+    // console.log(testFileDirectory);
+    // console.log()
+
+    // Parse each test js file.
+    testFileDirectory.forEach(fileName => { ParseTestJsFile(fileName); });
+
+    // Finalize our findings by re-writing out the TestList.js file ONLY if it has changed.
+    var index = testList_jsContents.indexOf('// ************ Generated Code Beyond this Point *************************');
+    var newFileContents =
+      '// ************ Generated Code Beyond this Point *************************\r\n' +
```
mattmazzola: Not required to change but you can use backtick (`) to create multiline strings. I find it more readable because there isn't extra characters in either side and definitely more editable.

```
const newFileContents = `
// ************ Generated Code Beyond this Point *************************
// Do NOT manually alter this file from this point onwards.
// Any changes you make will be overridden at runtime.
`

Only odd thing is it has to be left justified or the whitespace will be included.

File: cypress/GenerateCypressTestSpecs.js [hunk 12]
```diff
@@ -0,0 +1,121 @@
...
+    }
+    else console.log('NOTHING HAS CHANGED')
+  }
+  catch(error) { console.log(`Caught an Exception:\r\n${error}`)}
+}
+
+// ----------------------------------------------------------------------
+
+function ParseTestJsFile(fileName) {
+  var fileContents = fs.readFileSync(`${pathToTestFiles}/${fileName}`, 'ascii')
```
mattmazzola: Curios why we have to load as `ascii`? I would think `utf8` would work.

cypress/GenerateCypressTestSpecs.js
```diff
@@ -0,0 +1,147 @@
...
+  //console.log(`File ${fileName}:\r\n${fileContents}`)
+  
+  var fileLines = fileContents.split('\n')
+  
+  fileLines.forEach(line => {
+    line = line.trim()
+    if (line.startsWith('Cypress.TestCase(')) {
+      var iStart = line.indexOf("'", 17) + 1
```
mattmazzola: Would at least factor this out into a variable so we can get remove at least some of coupling and hard-coding.

```
const attributeName = `Cypress.TestCase(`
if (line.startsWith(attributeName)) {
      var iStart = line.indexOf("'", attributeName.length) + 1
...
```

In my opinion a better solution for this type of searching would be to combine use of RegEx so people can see what parts of the line your extracting by looking at the expression.

Doesn't account all the possibilities, which again is why this solution is brittle, but it's an example:

```
const attributeRegex = /Cypress.TestCase\('(\w+)',\s'([\s\w]+)'/
...
const matches = line.match(attributeRegex)
if (matches) {
   const groupName = matches[1]
   const testName = matches[2]
   ...
}
```

cypress/support/TestListManager.js
```diff
@@ -3,225 +3,143 @@
...
+  // function to a string containing the name of this function. It works
+  // like this:
+  //   1) We convert the function to a string containing the full function
+  //      that looks like: 
+  //        "function TheTestFunctionToRun(param1, param2) {...}"
+  //   2) The magic number 9 here is skipping past the string "function "
+  //      and grabbing just the name of the function.
+  var testFunctionAsString = testFunction.toString();
+  var testFunctionName = testFunctionAsString.substring(9, testFunctionAsString.indexOf('(', 9));
+  helpers.ConLog('TestListManager', `TestCase(${testGroupName}, ${testDescription}, ${testFunctionName})`);
```
mattmazzola: Similar to the comments I made about the `Cypress.TestCase` searching. Would at least extract the magic number to a variable and look at using RegEx.

E.g.
```
const functionPrefix = `function `
const testFunctionName = testFunctionAsString.substring(functionPrefix.length, testFunctionAsString.indexOf('(', functionPrefix.length))
```

It's better to code self document as much as possible. The comments could get stale, and we're optimizing for readability over performance.

cypress/GenerateCypressTestSpecs.js
SDESkowronski: EXCELLENT Suggestion! Keep em coming. (I missed that in the last CR)

SDESkowronski: I wasn't aware you could do that.

cypress/GenerateCypressTestSpecs.js
SDESkowronski: Isn't ascii the encoding vsCode is writing files out as?
I switched it to utf8 and tried it, but whenever I edit the file this code is using I see a message (generated by my test code) that the file had changed in the area of the file that I did not change, so I'll keep ascii unless there is a compelling reason to go to utf8.

cypress/support/TestListManager.js
SDESkowronski: Regex is overkill here, and not very easy to understand. Most people I know have to look up Regex strings to understand them.

I do agree with your comment about comments going stale, which is why I so rarely use them, but in this case I feel that this code benefits from the comment. I did however, fix the code to use the length as you suggested, that works for me.

cypress/GenerateCypressTestSpecs.js
SDESkowronski: I don't like to use Regex due to its complexity and difficulty to read, unless not using it results in significantly slower performance or the resulting code is significantly more complex than this code is. And I do have a lot of experience using Regex, its just that you use it once, then you don't use it for 2 or 3 months again. Difficult to maintain an understanding of it when used so rarely.

cypress/support/TestListManager.js
SDESkowronski: I made a note to self to review this code and fix it, but later not for this PR.

PR #954, opened 2019-01-24
chore: add cypress json schema reference

mattmazzola:
Minor PR just adding JSON Schema to the cypress file. Shouldn't affect any functionality.

See: https://docs.cypress.io/guides/tooling/intelligent-code-completion.html#Set-up-in-your-Dev-Environment-1

Before:
image

After:
image


mattmazzola:
Can one of you make a review with comment requesting changes so I can see how it behaves.

.vscode/settings.json
```diff
@@ -21,5 +21,13 @@
     ],
     "workbench.colorCustomizations": {
         "activityBar.background": "#164e86"
-    }
+    },
+    "json.schemas": [
+        {
+            "fileMatch": [
+                "cypress.json"
```
mattmazzola: Here is a comment

mattmazzola:
Hmm looks like the reviewer only gets a single "Resolve" button.  Cant do the pending stuff but it's better before when they didn't have any review system.

PR #958, opened 2019-01-25
fix: bug 1787 - show log dialog count (if triple digits)

pnwbrian:
DIsplay "99+" conditionally if log dialog count reaches 100


src/routes/Apps/App/Index.tsx
```diff
@@ -265,6 +266,7 @@ class Index extends React.Component<Props, ComponentState> {
                             </NavLink>
                             <NavLink className="cl-nav-link" data-testid="app-index-nav-link-log-dialogs" to={{ pathname: `${match.url}/logDialogs`, state: { app } }}>
                                 <OF.Icon iconName="List" /><span>Log Dialogs</span>
+                                <span className="count">{this.state.modelLoaded && (this.props.logDialogs.length > TRIPLE_DIGIT_LOGDIALOG_COUNT) ? `${TRIPLE_DIGIT_LOGDIALOG_COUNT}+` : ''}</span>
```
mattmazzola: Can you remind me what the deal was with using `logDialogs.length`?

I remember intentionally not showing it for some reason but can't remember why.  I'm wondering if the count is less than 99, why don't we just show the count?  

Like: 
```
... ? `${TRIPLE_DIGIT_LOGDIALOG_COUNT}+` : this.props.logDialogs.length
```

mattmazzola: Another way to think about this.  Based on the code you have it looks like the count area would be blank until they have more than 99 then it should `99+`.  This makes it still blank and then suddenly appear and not respond do any changes.

I think it's more reasonable for a user understanding our system to see the number go up and until 99 then just add + because the number is too large to display

pnwbrian: I remembered the collective decision apparently incorrectly. Minor change just pushed showing the count up to 99 then 99+, removing the blank display

LarsLiden: Not seeing it in the PR?

mattmazzola: > removing the blank display

Yea I'm not seeing it either.  This is marked as `resolved` but you still have the empty quotes on the end of the conditional which I think would make it a count with no value. Maybe you resolved before push on accident

pnwbrian: Github weirdness. I've just pushed the corrected fix and hand verified shows up in the GH webview. Fix was exactly as Matt suggested.

PR #962, opened 2019-01-28
 fix: scroll highlighted entity into view

mattmazzola:
Changes the `EntityPicker` in the `ExtractorResponseEditor` to scroll the highlighted item into view.
This allows people to see what is highlighted as they move the arrow keys.

Example:
http://recordit.co/SezKyKqNQO

Also fixes two other slightly related issues:
- Resets `highlightIndex` to 0 when the men is re-opened.  Previously if you select a word, scroll to the bottom, then deselect and select another word the `highlightIndex` would have stayed 

- Preserve border and indicate when no entities were found
http://recordit.co/YmLSbXGsBX

I plan to update the ActionPayloadEditor in a different PR.

I talked to @pnwbrian a bit about this earlier, but I want to review some changes that were made and the effects they're having. Maybe @LarsLiden can fill in gaps and answer.

Previously the menu was always as compact as it could be. Meaning if there were no entities matching the search text the menu would not be as tall.  This was chosen so minimize how much of the other parts of the background modal we are blocking with this modal.

There was a change to put the input/search box of the EntityPicker on top. The meant that the expanding list of items causes the place you were looking (the search input) to keep moving as you type which is bad. Then it was correct by forcing as fixed height which causes a larger then needed box.
**Why did we move the search box up and fix the height?**

The menu also used to hover over the word you selected. Now it is always positioned in the same spot even if the word you selected is on the far right.  I thought having the menu closer was a good for uses to associate the menu with the word by proximity.
**Why did we force the menu to always be on the left?**

Example:
image

I think it was much better when the menu was close to the selected word.


mattmazzola:
Talked with Lars offline.

Summary:
- Being fixed to left is a bug.
- Search input should be focused when a word is selected so user can type immediate it's also a bug.
- New Entity button should use same styling as other New Entity button in Actions modal
- Search input position at top and fixed height is not critical to design. We can try alternatives.

PR #964, opened 2019-01-29
style: entity creator cleanup

LarsLiden:
Cleaned up a few things:

- internationalization value was being used as a type rather than actual type
- isProgrammatic variable wasn't needed.  EntityType fully specifies this


src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -3,14 +3,15 @@
  * Licensed under the MIT License.
  */
 import * as React from 'react'
+import * as OF from 'office-ui-fabric-react'
+import * as Util from '../../../Utils/util'
```
mattmazzola: Not related do the review but this path looks weird. Why do have have two util? 
Especially since we're importing as wildcard I  wouldn't think it matters if we get all the stuff in `util` file plus extra stuff in `Utils` folder

PR #973, opened 2019-01-30
fix: bug 1897 - entity names cannot be pre-trained type names

pnwbrian:
Users could name their custom entities names like "number" and "email" before this change. We now establish some restricted names by borrowing the 'text' property of the dropdox options.

Name comparison is case-sensitive with datetimeV2 remaining blocked by the existing numeric check.


src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -160,18 +147,22 @@ class Container extends React.Component<Props, ComponentState> {
                     }))
 
             if (nextProps.entity === null) {
+                let reserved_names: string[] = []
                 const filteredPreBuiltOptions = localePreBuiltOptions.filter(entityOption => !nextProps.entities.some(e => !e.doNotMemorize && e.entityType === entityOption.key))
                 this.entityOptions = [...this.staticEntityOptions, ...filteredPreBuiltOptions]
                 this.resolverOptions = [...this.staticResolverOptions, ...localePreBuiltOptions]
 
+                filteredPreBuiltOptions.forEach(prebuilt_reserved_name => { reserved_names.push(prebuilt_reserved_name.text); })
```
mattmazzola: Minor but prefer to use the array methods instead of for loops or forEach.  It's more expressive and helps guarantee body of loop do unintended things.

Also like to keep the `camelCase` convention for variables. Not `prebuilt_reserved_name` -> `prebuiltReservedName`

Side note: I follow the Microsoft guidelines an only capitalize the first letter of acronyms. E.g. `HttpContext` instead of `HTTPContext`. I think it's easier to read and parse the words since you don't run into the capital letter of next word and often know what the acronym is anyways but Lars does all capital so choose what you like since we're already mixed here.

Also `prebuiltReservedName` isn't precisely true. The suffix `Name` sounds like it should be a string, but it's actually a `CLDropdownOption`.

Also would like keep the same concepts and variable names for the same thing. In one place you call it `restricted_entity_names` but in another you call it `reserved_names`. But are the same thing and should be called the same thing.

All that together means I would suggest:
- Remove initializer of empty array
- Change `restricted_entity_names` to `restrictedEntityNames`
- Change `reserved_names` to match `restrictedEntityNames`
- And change to `reserved_names = filteredPreBuiltOptions.map(preBuiltOption => preBuiltOption.text)`
- Use shorthand notation for assignment in object:
```
{
    ...
    restrictedEntityNames
}
```

pnwbrian: I'll probably adopt map() more if nothing else to promote array operations more in the code, so that's insightful. Camel casing is a mixed bag with Lars tending a slightly modified version with a heavier dose of capitalization.

The prebuildReservedName is really an array of strings as the .map like call is extracting only the .text property from the CLDropdownOption for subsequent use by .indexOf() array operator in the name validation code. I chose to extract here, otherwise the name validation code - which executes on keydown - would be more complex and lengthy, thereby impacting performance. Good catch on the assignment of objects shorthand, Initializing as empty is a holdover from my C/C++ days and empirical experience in explicit initialization catches more logic bugs and harnesses better JIT/compiler static analysis.

mattmazzola: After the talk offline I realize I was unclear when I said to remove the initializer.

This makes it sound like there will be no initial value and it will default to undefined. That is not what I meant. It's better to be as declarative as possible. THere is no reason for this varialbe to exist in a bad state (empty) and then later be populated. Just create it with it's initial value as the correct value.

pnwbrian: We landed on the same page after the talk. Next time in the code as these suggestions improve code quality

LarsLiden:
This isn’t the right fix.  It’s ok for a user to name an entity “number”, they just can’t name them “builtin-number”.  We intentionally prefix pre-builds with “buildin-“ so they can use them.

Essentially we just need to filter out any entity name that starts with “buildin-“ and were fine.

PR #977, opened 2019-01-31
feat: entity creation edit for enums

LarsLiden:
Note: save not implemented.  Waiting for server changes


LarsLiden:
re-trigger appveyor

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -368,7 +375,28 @@ class Container extends React.Component<Props, ComponentState> {
...
+        }
+
+        return ''
+    } 
+
+    isEnumDuplicate = (value: string): boolean => {
+        if (!value) {
+            return false
+        }
+        return (this.state.enumValues.filter(v => v === value).length > 1)
```
mattmazzola: Minor, `some` would be better here instead of `filter`.  It implies we only care if 1 of the condition is true once.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -25,7 +25,8 @@ const initState: ComponentState = {
     isPrebuilt: false,
     isMultivalueVal: false,
     isNegatableVal: false,
-    isEditing: false,
+    isEditing: false,  
+    enumValues: [undefined, undefined, undefined, undefined, undefined],
```
mattmazzola: Why did we have to use `undefined` instead of `''` ? It seems unnatural given we expect strings to be the only type of input to text field.

I'm assuming you put multiple entries here to show the user that they are expected to enter more than 1?

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
mattmazzola: Talked offline, `some` doesn't work in this case because `enumValues` already has the new value in it. So we're trying to detect 2 of them due to addition of new value, not if it existed in the array before.

PR #978, opened 2019-01-31
Bug Fix for Wait Non Wait Error Handling

SDESkowronski:
Also contains a fix for CreateModels.TagAndFrog


cypress/support/TestListManager.js
```diff
@@ -51,6 +51,7 @@ const zTemp = require('../tests/zTemp')
 // ----------------------------------------------------------------------
 
 function TestCase(testGroupName, testDescription, testFunction) {
+  throw 'Intentionally breaking all test cases'
```
mattmazzola: I expected this line to break the tests given the `throw` and the message but somehow the build on CircleCI passed when I expected it to fail.  Is this code not run or something?

SDESkowronski: Sorry, but that was meant to be a temporary change, I should have pinged you and let you know, I got busy and caught up in debugging...

At the end of my work day yesterday I published in this PR. I expected the tests to pass and move on. But when I looked this morning they were failing in essentially the same place. I made another fix expecting it to work, and nothing changed, so I doubted it was running my latest code, so I tested it by that alteration and I could see zero tests ran which proved it was running my latest code.

PR #980, opened 2019-01-31
chore: remove auth variables and react default files

mattmazzola:
Further cleanup to removes things we don't use. Don't want to confuse people thinking they are still needed.

- Removes the old authentication variables we used when we had AAD authentication. I forgot to remove them when we switched to LUIS key auth.  We can always look at older commit to reference how it was done if needed.
- Removes the default files that come with new react app.  We don't use the service worker or the logo file.
- Moves shared `NODE_OPTIONS` variable to the shared `.env` file.


mattmazzola:
From what I understand the the NODE_OPTIONS only affects the currently running node process.

Developers using our UI aren't running node. They are viewing the website in a browser.

Maybe @LarsLiden can clarify why this what added, but I'm guess it could only be an issue because of reacts's hot reloading build system.  I thought we only saw problems with the SDK running out of memory I can't remember what the issue was with this one.

Sometimes the process would fail with `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory` and I think this option prevents that by allowing node to use more memory.

LarsLiden:
That was a suggestion for the Appveyor customer support folks when our builds were hanging.

PR #981, opened 2019-01-31
fix: block builtin- prefix for custom,programmatic entities

pnwbrian:
Custom, enum and programmatic entities cannot have "builtin-" as prefix for their names so block attempts to name these with the "builtin-" if the name starts with this prefix


src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -359,8 +355,10 @@ class Container extends React.Component<Props, ComponentState> {
             return Util.formatMessageId(intl, FM.ENTITYCREATOREDITOR_FIELDERROR_ALPHANUMERIC)
         }
 
-        if (Object.values(this.state.restricted_entity_names).indexOf(value.toLocaleLowerCase()) >= 0) {
-            return Util.formatMessageId(intl, FM.ENTITYCREATOREDITOR_FIELDERROR_RESERVED)
+        if (!isNameUnrestricted) {
+            if ("builtin-" === value.toLocaleLowerCase().substring(0,"builtin-".length))  {
```
mattmazzola: Save the "builtin-" string as a variable and use it here and the `GetPrebuiltEntityName`.  The guarantees the prefix can't be out of sync.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -351,6 +345,8 @@ class Container extends React.Component<Props, ComponentState> {
     onGetNameErrorMessage = (value: string): string => {
         const { intl } = this.props
 
+        const isNameUnrestricted = this.state.entityTypeVal !== CLM.EntityType.LUIS && this.state.entityTypeVal !== CLM.EntityType.LOCAL && this.state.entityTypeVal !== CLM.EntityType.ENUM
```
mattmazzola: Why is this check needed? I thought we never wanted people to name an entity starting with `builtin-`.  Granted now that we don't have any logic based on the name, only the entityType, I think it should be Ok to do it but I wanted to know the reasoning?

If we must keep this check a few things to change:
1 - It would be more readable and editable to use a list and `includes`.  I think this would work:
```
const isNameUnrestricted = [CLM.EntityType.LUIS, CLM.EntityType.LOCAL, CLM.EntityType.ENUM].includes(this.state.entityTypeVal)
```  
Side node if you have this many boolean expressions it's better to put each one on it's own line like so:
```
const isNameUnrestricted = this.state.entityTypeVal !== CLM.EntityType.LUIS
    && this.state.entityTypeVal !== CLM.EntityType.LOCAL
    && this.state.entityTypeVal !== CLM.EntityType.ENUM
```
Again easier to read, edit, compare, etc...

2. This variable (`isNameUnrestricted`) should be located as closely to where it's used as possible.
In other words, don't declare it up here. Move it down to line `357`.  It's easier to people to see the whole .concept, otherwise they have to see a variable know it's not used and keep it in their head and then remember what it was later down the file.

mattmazzola: Oh I see, if the entity is a prebuilt then we must allow that prefix. Is that correct?

If so then maybe it's better to invert the logic and avoid listing of entity types which may change make the logic incorrect. We want to know when an entity is not a prebuilt and then apply the extra check.  Not assume if it's not those types it must be a prebuilt.  It's the same conversation we had the other day.  You want your logic to be precise and positive assertion when possible.

```
// Only apply builtin- check if entity is a prebuilt
if (typeof this.state.entityType !== 'string') {
    && value.toLowerCase().startsWith(EntityCreatorEditor.PrebuiltPrefix) {
  return Util.formatMessageId(intl, FM.ENTITYCREATOREDITOR_FIELDERROR_RESERVED)
}
```
Maybe we can use the helpers Lars added `isPrebuilt` although I'm not sure if we have the needed data/input type.

LarsLiden: I'd rather name this "isPrebuilt" as that explains exactly what is being calculated

pnwbrian: Threading the middle ground needle between reviewer comments

pnwbrian: Threading the middle ground needle between reviewer comments

LarsLiden:
If you are getting inconsistency between the review comments, please come talk to us so we can come to a consensus rather then try to guess how to accommodate

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -359,8 +356,10 @@ class Container extends React.Component<Props, ComponentState> {
             return Util.formatMessageId(intl, FM.ENTITYCREATOREDITOR_FIELDERROR_ALPHANUMERIC)
         }
 
-        if (Object.values(this.state.restricted_entity_names).indexOf(value.toLocaleLowerCase()) >= 0) {
-            return Util.formatMessageId(intl, FM.ENTITYCREATOREDITOR_FIELDERROR_RESERVED)
+        if (!isNameUnrestricted) {
```
LarsLiden: Please rename this to "isPrebuilt" it's more specific

PR #982, opened 2019-01-31
fix: bug 1896 - switching entity types needs to clear name field

pnwbrian:
Switching between entity types needs to clear the name field. Switching from pre-trained to custom for example left the pre-trained name linger in the name field. Now clear the name field, thereby requiring a new name be supplied.


src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -322,7 +322,7 @@ class Container extends React.Component<Props, ComponentState> {
             isMultivalueVal,
             isNegatableVal,
             entityTypeVal,
-            entityNameVal: isPrebuilt ? Container.GetPrebuiltEntityName(obj.text) : prevState.entityNameVal,
+            entityNameVal: isPrebuilt ? Container.GetPrebuiltEntityName(obj.text) : "",
```
LarsLiden: Seems like it would be a bit annoying to have is clear when it doesn't need to. I'd only clear it when switching away from a prebuilt

mattmazzola: Now that the new code detects the `builtin-` prefix I think when you switch away it will be an error anyways which would already force the user to change the name so maybe this isn't needed anymore. If we make it only clear on change from prebuilt it would make me wonder why it only clears the name some times.

LarsLiden:
Good point.  Let’s see what the experience is like.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
pnwbrian: Blanket clearing feels is the lesser annoying behavior vs conditionally clearing the field based on the previous drop selection. The latter seems inconsistently quirky.

LarsLiden:
So let’s try no clear with warning and see how that feels.

pnwbrian:
Reworked logic. Name field cleared if previous type was prebuilt if type is changed. Further, restrict names prefixed with "builtin-" as users enter the name. Third, preserve the name when possible.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -18,7 +18,7 @@ import { withRouter } from 'react-router-dom'
 import { RouteComponentProps } from 'react-router'
 import Component from './EntityCreatorComponent'
 
-const prebuiltPrefix = 'builtin-'
+const reservedNamePrefix = 'builtin-'
```
mattmazzola: I think it should remain as `prebuiltPrefix`. I'm not sure why github won't allow me to commnt on the line but see here:
https://github.com/Microsoft/ConversationLearner-UI/pull/982/files#diff-ecfa596a555464dcfc9af80d0fc3be98L62

I believe I said in a previous review, the constant should also be used in the `getPrebuiltEntityName` function to ensure these two stay in sync. 

Also when checking the error we use this check if the value starts with the prebuiltPrefix only when the input is not a prebuilt as these two are paired together.  A reserved name could be reserved for any reason and doesn't tell why.

pnwbrian: Synced both blocks of code to use the constant

PR #984, opened 2019-02-02
Feat: Enum support

LarsLiden:


src/components/ActionDetailsList.tsx
```diff
@@ -304,18 +357,7 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
...
             getSortValue: action => '',
-            render: (action, component) => action.requiredEntities.length === 0
-                ? <OF.Icon iconName="Remove" className="cl-icon" data-testid="action-details-empty-required-entities" />
-                : action.requiredEntities.map(entityId => {
-                    const entity = component.props.entities.find(e => e.entityId === entityId)
-                    return (
-                        <div className='ms-ListItem is-selectable ms-ListItem-primaryText' key={entityId} data-testid="action-details-required-entity">
-                            {entity
-                                ? entity.entityName
-                                : `Error - Entity ID: ${entityId}`}
```
mattmazzola: Can you remind me in which cases we will have an an `entityId` but no `entity`?
Only one I could think of is because we allow deleting an entity even if it's in use.  All the actions and train dialogs that referenced it become invalid and show this error?

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -67,12 +67,72 @@ const convertEntityIdsToTags = (ids: string[], entities: CLM.EntityBase[]): OF.I
         .map<OF.ITag>(convertEntityToTag)
 }
 
-const getSuggestedTags = (filterText: string, allTags: OF.ITag[], tagsToExclude: OF.ITag[]): OF.ITag[] => {
+// Entities that can be chosen for required / blocking
+const conditionalEntityTags = (entities: CLM.EntityBase[]): IConditionalTag[] =>  {
+
+    // Ignore resolvers and negative entities
+    const filteredEntities = entities.filter(e => !e.doNotMemorize && !e.positiveId)
```
mattmazzola: I think this `doNotMemorize` term was added while I was gone. What does `doNotMemorize` mean? Or what's it used for?

LarsLiden:
When we create a prebuilt to use for an entity with a resolution and the prebuilt also doesn’t exist on its own, we don’t want to set the memory when the prebuilt is detected by LUIS

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -67,12 +67,72 @@ const convertEntityIdsToTags = (ids: string[], entities: CLM.EntityBase[]): OF.I
...
+
+const getSuggestedTags = (filterText: string, allTags: OF.ITag[], tagsToExclude: OF.ITag[], mutuallyExclusive: OF.ITag[] = []): OF.ITag[] => {
     let fText = (filterText.startsWith(ActionPayloadEditor.triggerCharacter) ? filterText.substring(1) : filterText).trim()
 
-    const availableTags = allTags
+    let availableTags = allTags
         .filter(tag => !tagsToExclude.some(t => t.key === tag.key))
 
+    // Check for mutually exclusive conditions and remove them 
+    availableTags = availableTags
```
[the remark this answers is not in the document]
LarsLiden: Make it easier for me to debug.  Could be chained

src/components/ActionDetailsList.tsx
LarsLiden: With the most recent version we try to let users more or less do whatever they want in terms of editing, with warnings and errors when they do so.  Also, if multiple people are editing the same model one could end up a weird state.  Generally trying to detect all these cases and handle them gracefully.

PR #985, opened 2019-02-02
feat: force package creation

LarsLiden:
Previous check-in used "Feat".  It's case sensitive and didn't push a new package

### DIFF
--- a/src/components/modals/EntityExtractor.tsx
+++ b/src/components/modals/EntityExtractor.tsx
@@ -3,23 +3,23 @@
  * Licensed under the MIT License.
  */
 import * as React from 'react'
+import * as Util from '../../Utils/util'
+import * as CLM from '@conversationlearner/models'
+import * as OF from 'office-ui-fabric-react'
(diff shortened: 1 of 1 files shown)


src/components/modals/EntityExtractor.tsx
```diff
@@ -3,23 +3,23 @@
  * Licensed under the MIT License.
  */
 import * as React from 'react'
+import * as Util from '../../Utils/util'
+import * as CLM from '@conversationlearner/models'
+import * as OF from 'office-ui-fabric-react'
+import * as ToolTips from '../ToolTips/ToolTips'
+import * as ExtractorResponseEditor from '../ExtractorResponseEditor'
```
mattmazzola: I noticed this in other reviews. Is this just because you like wildcard imports at top?

I think there are some extensions or Prettier has import sorting options that way you don't have to do this manually if you are.

LarsLiden:
Yes.  I’ve just been opportunistically cleaning stuff up when I make changes to imports.  Makes it easier to read

PR #986, opened 2019-02-04
Persistent logging

SDESkowronski:
Code to persist the console logs to file so that debugging CircleCI test failures is easier.


cypress/support/PersistentLogs.js
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+ */
+
+var testCaseRegistry = [];
+var indexRegistry = 0;
+var logEntries = '';
+
+// Override the original console.log() function
+var originalConsolLog = console.log;
```
pnwbrian: I'd recommend "const" over "var" to avoid accidental hijacking elsewhere in the code

cypress/support/PersistentLogs.js
```diff
@@ -0,0 +1,31 @@
...
+Cypress.PersistentLogs.RegisterTestCase = (testGroupName, testName) => {
+  testCaseRegistry.push(`${testGroupName}/${testName}`);
+}
+
+// After each test case ends...
+afterEach(() => {
+  var logFileName = `./results/cypress/${testCaseRegistry[indexRegistry++]}.${Cypress.moment().format("YY.MM.DD.HH.mm.ss..SSS")}.log`;
+  cy.writeFile(logFileName, logEntries);  
+  logEntries = '';
+});
```
pnwbrian: Does Cypress offer a async/non-blocking file operation. I ask given the suspected timing issues we've encountered in the past.

cypress/support/PersistentLogs.js
SDESkowronski: Done.

mattmazzola: Is this pushed? There's a bunch of things that still use `var`.

cypress/support/PersistentLogs.js
```diff
@@ -0,0 +1,31 @@
...
+
+var testCaseRegistry = [];
+var indexRegistry = 0;
+var logEntries = '';
+
+// Override the original console.log() function
+var originalConsolLog = console.log;
+
+// This becomes the effective console.log() function.
+console.log = function (message) {
```
mattmazzola: Does cypress offer any built in features that can save stdout as a log file? 

It's generally recommended to avoid overriding browser defaults and instead use a custom function so people expect it will behave differently and are not misled. I though you were already using a custom `helpers.conLog` everywhere? Can we just extend that one do write add to the file instead?

I guess in this case it ensures people can't use the regular log and forget to use the helper and mistakenly avoid writing the file.

cypress/support/PersistentLogs.js
SDESkowronski: I believe this is async because you can chain a .then() function off of it, they offer no other version. But you did bring up a good point, I should clear those logEntries in the chained .then() function...not even sure why it worked that way but it did.

I fixed the code.

cypress/support/PersistentLogs.js
SDESkowronski: Well Cypress does have plans of including this feature in their code, but it is not ready yet. I discovered that just yesterday while researching another issue. **So when theirs is released we can just drop this file and everything will continue to work.** Which is one good reason not to change our helper method.

cypress/support/PersistentLogs.js
```diff
@@ -0,0 +1,31 @@
...
+  originalConsolLog.apply(console, arguments);
+  logEntries += message + '\r\n';
+};
+
+// This will be called by the test list manager to register each test case
+// in the order the tests will be run in.
+Cypress.PersistentLogs = {};
```
mattmazzola: I noticed this pattern in the other reviews. Why is it needed?

It seems it's only there to share functions defined here with the test list manager. Is that true or does it serve some other purpose?
If so it seems similar to writing global variables on the browser `window` object and being able to reference them anywhere.  Similarly with concept of over writing of browser defaults or using global variables it's generally avoided.

Are there any alternatives to sharing the function?

cypress/support/PersistentLogs.js
```diff
@@ -0,0 +1,30 @@
...
+
+var testCaseRegistry = [];
+var indexRegistry = 0;
+var logEntries = '';
+
+// Override the original console.log() function
+const originalConsolLog = console.log;
+
+// This becomes the effective console.log() function.
+console.log = function (message) {
```
mattmazzola: Similarly all the vars can be `let`.  It generally goes like this. Use `const` everywhere by default and only change to `let` when you realize you need to re assign it. Shouldn't ever need to use `var` as it has all the javascript hoisting quirkiness and this makes it `undefined` in certain scenarios or available to used out of intended scope which aren't caught. `let` and `const` prevent these mistakes.

cypress/support/PersistentLogs.js
mattmazzola: Well the normal the alternative to global variables is to actually define a module and `import` or `require` it so it's an explicit dependency.  I think in a previous review you said this doesn't work but I couldn't remember. It may not be an option but I wanted ask because we need to try to limit this pattern.  It seems like we found global variables is working for what you needed and the more we add them we just contribute to the tech debt we eventually want to get away from.

cypress/support/TestListManager.js
```diff
@@ -142,5 +143,5 @@ function GetTestGroup(name) {
 }
 
 function GetTest(testGroup, testNameToFind) {
```
mattmazzola: Still think it's weird to have what looks like a `Get` method internally be a `find`.  Do we expect this method to ever not find a test?  If not, then I would change this to `throw` an error if `find` returns `undefined`

Like:
```

cypress/support/PersistentLogs.js
SDESkowronski: I was not aware of these distinctions, so **thanks for pointing them out**. I did a little looking online and I came across this article which I agree with: [JavaScript ES6+: var, let, or const?](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75) so moving forward I will adopt this style.

But for the record, I disagree with the inventors of JavaScript including the `let` keyword in the language, I don't think it is necessary. However, since it is the norm, I am willing to go along to get along.

cypress/support/TestListManager.js
SDESkowronski: I've never really considered the subtle difference between a GET and FIND method, I've always lumped them into the same category. So again I thank you for continuing to educate me on such things.

I changed those two methods (GetTestGroup too) to reflect this way of thinking about it.

mattmazzola: Hmm, I had commented on this previously here and you had even replied so I thought I would bring it up again.
https://github.com/Microsoft/ConversationLearner-UI/pull/953#discussion_r251109416

Thanks for changing it. Even if you don't agree, javascript is written with thinking so want to keep consistent.

cypress/support/PersistentLogs.js
mattmazzola: Thanks for changing and thanks for replying on GitHub instead of email.

It's a bit ironic that the higher level point of the article you linked to is:
> keep things simple.

Variable types is only one small part of this. It reinforces the reason why I expressed concern over the complexity of test list manager, spec generator and such. They are not simple.  Since I can't offer alternatives it's best to go ahead but just keep this in mind.

Anyways at least the article helped explain the variable situation. Thanks

cypress/support/PersistentLogs.js
SDESkowronski: The `import`/`require` model you suggest does not work in this case. 

I believe I have used this technique sparingly, only when necessary, but as I continue to work with the code I will keep it in mind and remove any instances of it that are not necessary.

PR #987, opened 2019-02-04
Fix Generate Cypress Test Specs

SDESkowronski:
I was using the back-tick quote on the generated code which omitted the Carriage Return character from the file. Whenever a user hand modified the TestList.js file, but not anything in the masterListOfAllTestCases it would re-generate the file unnecessarily. This creates unnecessary noise in the output and unnecessary need to push in a code change that is not really a change.

### DIFF
--- a/cypress/GenerateCypressTestSpecs.js
+++ b/cypress/GenerateCypressTestSpecs.js
@@ -58,7 +58,7 @@ function main() {
 
     // Get the test list file that we MIGHT need to modify.
     const pathToTestList = './cypress/TestLists.js';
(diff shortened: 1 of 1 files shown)


SDESkowronski:
Yes I confirmed the backtick wasn’t generating carriage returns, don’t know why, I thought the same as you it should know the OS and work appropriately. Definitely NOT a side effect of my code.

Why it matters is when a developer edits the file, which we do quite often when working with tests, the editor puts in the carriage returns…

  *   then the generator detects the change and rewrites the file
  *   then GitHub detects the file as changed and wants to commit the file

PR #989, opened 2019-02-05
feat: add tags and description to train dialogs

mattmazzola:
Currently not working correct due to the matrix of edits allowed but I wanted to show what I have to get feedback.


src/components/BorderlessInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+    active: boolean
+}
+
+class component extends React.Component<Props> {
+    state: State = {
+        active: false
+    }
+
+    onFocusInput = () => {
+        console.log(`onClickInput`)
```
LarsLiden: left this in on purpose?

src/components/BorderlessInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+        this.setState({
+            active: true
+        })
+    }
+
+    onBlurInput = () => {
+        console.log(`onBlur`)
```
LarsLiden: left this in on purpose?

src/components/BorderlessInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+        this.setState({
+            active: false
+        })
+    }
+
+    onChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const inputValue = event.target.value
+        this.props.onChange(inputValue)
```
LarsLiden: any reason to add the const?  Could just set directly?

src/components/Tags.tsx
```diff
@@ -0,0 +1,88 @@
...
+        event.preventDefault()
+        const tag = this.state.inputValue
+        this.props.onAdd(tag)
+        this.setState({
+            inputValue: ''
+        })
+    }
+
+    onChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const inputValue = event.target.value
```
LarsLiden: is const necessary?

LarsLiden: Consider doing a toUpperCase() here so that it's always upper case

src/components/TagsDisplay.tsx
```diff
@@ -0,0 +1,26 @@
...
+    tags: string[]
+}
+
+class component extends React.Component<Props> {
+    render() {
+        const { tags } = this.props
+        return (
+            <div className={`cl-tags-display ${this.props.className}`}>
+                {tags.length === 0
+                    ? `0 tags`
```
LarsLiden: Perhaps "No Tags" is more user friendly?

src/components/modals/EditDialogAdmin.tsx
```diff
@@ -376,23 +378,37 @@ class EditDialogAdmin extends React.Component<Props, ComponentState> {
...
+                            />
                         </div>
-                        <div className="cl-ux-flexpanel--right" style={{ width: "20%", marginRight: '3em' }}>
-                            <TrainingStatusContainer
-                                app={this.props.app}
+                        <div className="cl-field">
+                            <label htmlFor="description">Description:</label>
+                            <BorderlessInput
+                                id="description"
+                                placeholder="Click on to add description"
```
LarsLiden: Will need to internationalize at some point

src/components/BorderlessInput.tsx
mattmazzola: Set what?
Did you mean this: `this.props.onChange(event.target.value)`

If so, I will change it. For some reason I thought it made it more clear by adding a name for what the `event.target.value` is but on second look it seems obvious since we have the React generic  `HTMLInputElement` so it must be an input value.

src/components/BorderlessInput.tsx
mattmazzola: Nope, will remove, pushed this early to get feedback on the change detection or save type issue an should have cleaned it up first.

src/components/BorderlessInput.tsx
LarsLiden: yeah, that's what I meant

src/components/Tags.tsx
mattmazzola: I don't think we should modify the user's input. I think they should feel in control and we should let them do what they want. We already have this philosophy with being able to edit the dialogs.  We let them edit entities in use and invalidate things because it's what they want to do. 

The text search and comparison is still case insensitive so you can still find tags.

src/components/TagsDisplay.tsx
mattmazzola: Yea, changed. I see what people mean by being confused about GitHub not showing the updated file on the review comment. Here it is: https://github.com/Microsoft/ConversationLearner-UI/pull/989/files#diff-c19a8069a2276942e428e7b1b27f112bR14

I think there actually are two different cases for the text and context is only known by the parent. I didn't implement it yet but I will if you think it's worth it.

The `<TagsReadOnly />` component my be used in various places and when there are no tags given it needs to still render something useful like `No Tags` to the user. However, on the `DetailsList` then I think it should say something similar to the "description" column like: "Open dialog to add tags"

src/components/modals/TeachSessionModal.tsx
```diff
@@ -155,6 +173,22 @@ class TeachModal extends React.Component<Props, ComponentState> {
                 replaceActivityIndex
             })
         }
+
+        // If we just added a sourceTrainDialog it's because we continued from an existing Train Dialog
+        // Or if the tags or description were changed before continuing
+        // Copy over the tags and description from it
+        if (!this.props.sourceTrainDialog && newProps.sourceTrainDialog
```
pnwbrian: Logic conditions a bit hard to follow and validate; suggest more liberal use of parens, or reworking logic if possible

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -102,51 +90,25 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
...
-                        {firstInput}
-                    </span>)
-                }
-                return <OF.Icon iconName="Remove" className="notFoundIcon" />
-            },
-            getSortValue: trainDialog => {
-                let firstInput = getFirstInput(trainDialog)
-                return firstInput ? firstInput.toLowerCase() : ''
-            }
+            render: trainDialog => <TagsReadOnly className={textClassName(trainDialog)} tags={trainDialog.tags} />,
```
mattmazzola: @SDESkowronski 

These changes would break the tests. The `data-testid="train-dialogs-first-input"` which is used in `GetFirstInputs` and `ClickTraining`. I think changing `ClickTraining` to use updated `data-testid="train-dialogs-tags"` is fine since it doesn't matter which column we click, but the things like `EditTraining` depend on knowing firstInput which is no longer there.

For now I think the easiest solution is to add the `data-testid="train-dialogs-first-input"` and `data-testid="train-dialogs-last-input"` so the tests can still find and make use that information.

I see most of the tests rely on knowing the `firstInput` `lastInput` and `lastResponse` to interact with the train dialogs. Is there another way to know which row to click on?

I'm thinking in the future each train dialog created by the tests could add a specific tag or description such as `createdByTest` or `e2etest-branching`. Then search for that tag or description. This would allow finding specific dialogs to tests even if they have the same inputs and response fields.

mattmazzola:
> now saving on detected changes

Can you explain this a bit more? Like which line? Might be related to one of the bugs I saw.
I had not intended to save on change. I think the user should always have to click the button and only then does it save.

> change two product elements

Also confused by this. Which elements? And why only two?
It changes the server, models, ui, and tests. Yes it's a large change but you're suppose to keep the PR's as atomic things like entire features are entire bugs. That way people can revert if needed.  If you commit  half a feature it gets harder to manage and I think it's harder to review since you can't see all the code changes in one place and have to reference back and forth.

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -102,51 +102,34 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
...
-                if (lastInput) {
-                    return <span className={textClassName(trainDialog)} data-testid="train-dialogs-last-input">{lastInput}</span>
-                }
-                return <OF.Icon iconName="Remove" className="notFoundIcon" />
+                const firstInput = getFirstInput(trainDialog)
+                const lastInput = getLastInput(trainDialog)
+                return <>
+                    <span data-testid="train-dialogs-description" className={textClassName(trainDialog)}>{trainDialog.description || `Open dialog to add description`}</span>
+                    {/* Keep firstInput and lastInput here until tests are upgraded */}
+                    <span style={{ visibility: "hidden", position: "absolute" }} data-testid="train-dialogs-first-input">{firstInput ? firstInput : ''}</span>
```
mattmazzola: @SDESkowronski 

My first comment somehow go attached to review and is now stale, but here is what I was suggesting.  Preserves the content and `data-testid` attributes in hidden elements.

src/components/modals/EditDialogAdmin.tsx
mattmazzola: Yea, added internationalization for both components.

cypress/integration/tagsAndDescription.spec.js
```diff
@@ -0,0 +1,116 @@
+describe('Tags and Description', () => {
+    context('Train Dialogs', () => {
```
mattmazzola: @pnwbrian @SDESkowronski 

Here is outline of tests I think are relevant.  An alternative test categorization is to update all the existing tests with the tags and description stuff integrated within them. I could see it both ways but thought it was clear we are testing tags and description feature when its separated like this.

How do we make this format of writing tests work in the manager?
It seemed like the current code is only allowing a test group and name, and requires repeating the group and name for each test. I think it should be arbitrary and up to the test creator to decide how many levels make sense for those tests.  All our unit tests are written this way also.

End result to look something like this:
image

mattmazzola: I ran the `cypress:dev` command you added and tried to run a train related test to see why the First Inputs are failing but it is not working.  I don't know how it's possible but the top level URL is changing to `localhost:5050` instead of the cypress frame url. 

See: http://recordit.co/2rurbLKsz3

Why is this happening?

I even tried modifying the `testList` array and running `(RunTestsFromList)` and it still didn't work.

mattmazzola:
Since this is hard to capture in pictures took a video:
https://microsoft.sharepoint.com/teams/convsysresearch/Shared%20Documents/Forms/AllItems.aspx?slrid=41efb69e-e0cf-0000-b690-368c22799cc7&viewid=a70ab912-6a5d-4369-bda7-749835f301d4&id=%2Fteams%2Fconvsysresearch%2FShared%20Documents%2FGeneral%2FtagsAndDescriptionDemo%2Emp4&parent=%2Fteams%2Fconvsysresearch%2FShared%20Documents%2FGeneral

mattmazzola:
Previously:
image

There was suggestion and discussion to change "Description" to "Scenario" and move it to the top.
image

I think description is the most open term and makes the user think they can put whatever they want in there. The other suggestion was title but I thought that was to restrictive and not the type of information we wanted to capture.  Scenario seems like a middle ground.

In my mind tags would be the higher level concept or scenario that they want to categorize on and the description is more specific of what actually happened in the dialog.  I'm sure we'll hear from customers either way.

src/components/TagsDisplay.tsx
mattmazzola: I added it in latest push

cypress/integration/tagsAndDescription.spec.js
SDESkowronski: Did you resolve this? I answered your question about this in teams last week. If you merge in from Master you will have the fix.

mattmazzola: Yes, I saw that comment. Tests are running now. The ones that seem to be related to train dialogs which this PR would affect are all passing:

image

Oddly enough the tests that are failing are for model creation:

> Error: the string "Exact Match 'TEXT' NOT Found" was thrown, throw an Error :)

Which is coming from `ExactMatch` failure in this:
```
export function SelectType(type)
{
  cy.Get('[data-testid="dropdown-action-type"]').Click()
  cy.Get('button.ms-Dropdown-item').ExactMatch(type).Click()
}
```

Which this PR shouldn't have anything effect on.

src/routes/Apps/App/TrainDialogs.tsx
SDESkowronski: You did good on this change! I got the branch and tried a couple of my tests that rely on this and they work...but then you probably already knew that.

I'm not sure that we will want to rely on this long term, but it is a good way to bridge the gap.

cypress/integration/tagsAndDescription.spec.js
mattmazzola: I think more importantly is the question:
> How do we make this format of writing tests work in the manager?

Certainly out of scope for PR but it is a goal for the long term and something I want you to think about.

SDESkowronski: As for the way you have organized the test cases...

There is a lot of merit in your proposal, specifically I can see that the test list clearly documents what is covered by the test scenarios. However, the realities of test automation will likely restrain and curtail your goals.

Setup of the test case takes a significant amount of time, which adds up when spread across many test cases. Along the way many other aspects of the UI under test are exercised as well, so in reality you are really testing many of the features all at once. If we created the tests you describe in the image above it would take approximately 7.5 minutes to run v.s. 3.5 minutes if those scenarios were all tested in 2 or 3 test cases.

Also there is the extra time it take to carefully craft all of the extra test cases to focus on a single specific feature and scenario...maintenance costs too.

The way I have organized the tests are to take advantage of this reality and I test and validate (nearly) every step I take towards the end goal. There are far fewer individual test cases, yet they all test a lot of features. 

This by the way is how the Cypress documentation says it should be done as well. I was happy to see that in their docs since this has been my issue with test automation that I had to maintain on other teams I was on. There were way too many tests to maintain because it was organized more like you are suggesting. There was a lot of duplicate effort. Although they did use common helper methods, still there was a lot more code created to test than was really necessary.

mattmazzola: > This by the way is how the Cypress documentation says it should be done as well.

Yes, I think there are good reasons for it.

> Setup of the test case takes a significant amount of time  ... it would take approximately 7.5 minutes

What contributes to this time?  I thought in most cases we import a model which means there isn't any setup since it's pre-configured. I thought after training completes it would be ready to be tested?
If importing is very slow we could get by with creating a new model for this test since it only needs 1 action to select. From there it could be same train dialog for the remainder.  We don't have to re-import or create a model for each of those lines only once.

> Along the way many other aspects of the UI under test
> I test and validate (nearly) every step I take towards the end goal

The goal for these tests to verify tags and description. If we tests everything all the time this seems like it would take a long time and make the tests brittle. They could break due to unrelated changes which mislead people.  I would think the test should be isolated and specific so we know where the problem is when it occurs.  E.g. If I write a test for tags then later change how the tags work I would expect that test and only that to fail so I know which one to fix. It would not affect tests related to model creation, entities, or settings and other features.

> There were way too many tests to maintain
> There was a lot of duplicate effort.

I think it's about maximizing quality of product and developer productivity rather than minimizing test count.  It doesn't matter if there are 100's of tests if I only have to fix 1 that I broke and look in one location.  

I don't understand how there is duplicate effort.  I would think this categorization of tests means less duplicated effort because it doesn't spend time finding and asserting things that have already been asserted in previous tests.  

> I can see that the test list clearly documents what is covered by the test scenarios

Yes, that is very important and something current tests are missing.  Without looking at how the tests are implemented someone should be able to see what it's testing and asserting. Also when a test fails it will show an 'X' next to the tests that failed instead of you having to log every error and recreate a reporting system.

Example we just came across:
image

It only gives you the test name, not the specific failure. I would expect this to tell something like "should select the action type TEXT". If someone doesn't know where "TEXT" is written in the code or know it's a type of action they would not know where to look for this message. It could seemingly come from any where in the test.

SDESkowronski: With regards to this issue of:
> Error: the string "Exact Match 'TEXT' NOT Found" was thrown, throw an Error :)

This is due to the ELECTRON browser being used by CircleCI runs...the .innerText() function off of the elements has a line feed character in it when the Chrome browser does not.

SDESkowronski:
There is a bug in this PR. Repro Steps:

1. Create a new Train Dialog and save it.
2. Edit the Train Dialog and modify either "Scenario" or "Tags"

**Expected:** The `Abandon` button to show up.
**Actual:** `Save Edit` or `Delete` buttons are the only way out of the dialog.

mattmazzola:
Merging bit early to unblock Lars.  I had mistakenly merged the corresponding change in models before this was also ready to be merged.

PR #992, opened 2019-02-11
Code Cleanup var let Semicolons Removed Litterals Added

SDESkowronski:
Cleaned up code according to my understanding of what Matt said were JavaScript norms and standards.

### DIFF
--- a/cypress/support/components/EditDialogModal.js
+++ b/cypress/support/components/EditDialogModal.js
@@ -64,8 +64,8 @@ export function SelectChatTurn(message, index = 0) {
         if (index > 0) index--
         else {
           helpers.ConLog(funcName, `FOUND!`)
-          elements[i].click();
-          return i;
+          elements[i].click()
+          return i
         }
       }
       else helpers.ConLog(funcName, `NOT A MATCH`)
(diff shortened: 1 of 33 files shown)


cypress/support/components/EditDialogModal.js
```diff
@@ -131,26 +131,26 @@ export function VerifyThereAreNoChatEditControls(userMessage, botMessage) {
...
-      var elements = Cypress.$('[data-testid="token-node-entity-value"] > span > span');
+      var found = false
+      var elements = Cypress.$('[data-testid="token-node-entity-value"] > span > span')
 
       // If you need to find a phrase, this part of the code will fail, 
       // you will need to upgrade this code in that case.
       var element = elements.find(element => element.innerText === text)
       if (element) {
-        found = Cypress.$(element).parents('.cl-entity-node--custom').find(`[data-testid="custom-entity-name-button"]:contains('${entity}')`).length == 0;
+        found = Cypress.$(element).parents('.cl-entity-node--custom').find(`[data-testid="custom-entity-name-button"]:contains('${entity}')`).length == 0
```
mattmazzola: Don't need to change but try to use strict equals (`===`) whenever possible. It helps avoid the implicit type coercion that javascript does and is often what you intend to check. There are some cases you want type coercsion like checking for `null` && `undefined` using a single `==` but it's rare.

Don't think applies in this scenario since `find` probably guarantees to return an array but the canonical examples are:
```
undefined == null
// true
false == 0
// true
```
These cases could lead to unintended behavior of the app.

SDESkowronski: Okay, got it.

PR #993, opened 2019-02-12
Fix cy.DoesNotContain

SDESkowronski:
While beginning work on another test case I noticed that this method looked wrong. Because this method passes when an element does not exist on the page, it is easy to get it wrong and not notice. So I created a test case to verify it, and sure enough the code in it was bad (my bad). Now we can keep an eye on it if we ever have to change it and make sure it keeps working.

### DIFF
--- a/cypress/support/MonitorDocumentChanges.js
+++ b/cypress/support/MonitorDocumentChanges.js
@@ -162,8 +162,7 @@ var MonitorDocumentChanges = (function () {
     var currentHtml = Cypress.$('html')[0].outerHTML
     if (currentHtml != lastHtml) {
       helpers.ConLog(thisFuncName, `Change Found - Milliseconds since last change: ${(currentTime - lastChangeTime)}`)
-      if (dumpHtml || expectingSpinner)
-        helpers.ConLog(thisFuncName, `Current HTML:\n${currentHtml}`)
+      if (dumpHtml) { helpers.ConLog(thisFuncName, `Current HTML:\n${currentHtml}`); }
 
       lastChangeTime = currentTime
       lastHtml = currentHtml
@@ -189,7 +188,7 @@ var MonitorDocumentChanges = (function () {
         SetExpectingSpinner(false)
 
         if (spinnerTexts[i] != currentSpinnerText) {
-          helpers.ConLog(thisFuncName, `Start - ${spinnerTexts[i]} - current HTML:\n${lastHtml}`)
+          if (dumpHtml) { helpers.ConLog(thisFuncName, `Start - ${spinnerTexts[i]} - current HTML:\n${lastHtml}`); }
           currentSpinnerText = spinnerTexts[i]
         }
         return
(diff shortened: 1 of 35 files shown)


cypress/support/MonitorDocumentChanges.js
```diff
@@ -68,31 +68,34 @@ var MonitorDocumentChanges = (function () {
...
 
-        helpers.ConLog(`cy.DoesNotContain()`, `Found ${elements.length} for selector: ${selector}`)
+        helpers.ConLog(functionSignature, `Found ${elements.length} for selector: ${selector}`)
 
-        if (elements.length > 0) {
-          if (arg2) {
-            elements = Cypress.$(elements).find(`:contains(${arg2})`)
-            helpers.ConLog(`cy.DoesNotContain()`, `Found ${elements.length} containing text: ${arg2}`)
-          }
+        if (elements.length > 0 && textItShouldNotContain) {
```
pnwbrian: Suggest wrapping subexpression in parens for readability

cypress/support/MonitorDocumentChanges.js
```diff
@@ -68,31 +68,34 @@ var MonitorDocumentChanges = (function () {
...
-            throw `Selector "${arg1}" + "${arg2}" was expected to be missing from the DOM, instead we found ${elements.length} instances of it.`
+        if (elements.length > 0) {
+          if (expectFailure) return true;
+          throw `selector: "${selector}" & textItShouldNotContain: "${textItShouldNotContain}" was expected to be missing from the DOM, instead we found ${elements.length} instances of it.`
         }
-        helpers.ConLog(`cy.DoesNotContain()`, `PASSED - Selector: ${selector} was NOT Found as Expected`)
+        helpers.ConLog(functionSignature, `PASSED - Selector was NOT Found as Expected`)
+        if (expectFailure) return false;
       })
     })
```
pnwbrian: If elements.length evaluates to zero length, and expectFailure is false (default), the code's return value is non-deterministic. Suggest considering a default initializer to textItShouldNotContain

cypress/support/MonitorDocumentChanges.js
SDESkowronski: ok

cypress/support/MonitorDocumentChanges.js
SDESkowronski: > Suggest considering a default initializer to textItShouldNotContain

Can you give me a reason why. My only resistance is that I don't see the need so less code is less noise.

SDESkowronski: > If elements.length evaluates to zero length, and expectFailure is false (default), the code's return value is non-deterministic.

This is intentional.

When `expectFailure` is true, that is test mode. I need the return value to test it. I cannot use try/catch since Cypress catches all exceptions and stops the test.

When `expectFailure` is false, that is the normal run mode. I need it to return the value that came from the optional previous cy command that this may be chained off of in order to do further work. So we could code something like this for example:

    `cy.Get('.selector').DoesNotContain('.notSelector').Click()`

So in that example, the DoesNotContain would be an intermediate validation step in the chain, and the click would click on the first element returned from the Get(`.selector`).

PR #994, opened 2019-02-14
Cannot Create Non-Wait END_SESSION Action

SDESkowronski:
This adds test coverage to an existing test to ensure that bug 1910 is fixed in regards to unchecking the `Wait for Response` field before selecting Action type `END_SESSION` when creating a new Action.

The single line of code that enforces this rule has been commented out, doomed to fail once the bug is fixed, that failure will signal me to remove the comment and activate that line of code.

ALSO this code is a part of creating complete code coverage for a scenario that will verify we cannot create a training with two `END_SESSION` action types in it.


cypress/support/Actions.js
```diff
@@ -16,24 +16,35 @@ const helpers = require('../support/Helpers')
...
   actionModal.ClickNewAction()
-  // TODO: this is the default but we need to get this working... actionsModal.selectTypeText()
-  actionModal.TypeResponse(response)
-  actionModal.TypeExpectedEntity(expectedEntities)
-  actionModal.TypeRequiredEntities(requiredEntities)
-  actionModal.TypeDisqualifyingEntities(disqualifyingEntities)
+
+  // We do this first since we had a bug (1910) where it is not reset by the UI when
+  // type END_SESSION is selected.
   if (uncheckWaitForResponse) actionModal.UncheckWaitForResponse()
```
mattmazzola: I thought unchecking the wait before you select the END_SESSION type was the bug. This only works by taking advantage of that bug right?  I would think this should be placed after the `SelectType` and the test should fail if they attempt to uncheck an END_SESSION action.

cypress/tests/CreateModels.js
```diff
@@ -52,6 +53,7 @@ export function WaitVsNoWaitActions()
   actions.CreateNewAction({ response: 'Cows say moo!', uncheckWaitForResponse: true })
   actions.CreateNewAction({ response: 'Ducks say quack!', uncheckWaitForResponse: true })
   actions.CreateNewAction({ response: 'Fish just swim.', uncheckWaitForResponse: true })
+  actions.CreateNewAction({ type: 'END_SESSION', response: "That's All Folks.", uncheckWaitForResponse: true })
```
mattmazzola: I didn't think this should be possible.  I would expect it to fail when given those parameters.

Expecting: `uncheckWaitForResponse: false`

cypress/support/Actions.js
SDESkowronski: You are right that this is exercising that bug. I am intentionally doing the wrong thing so as to produce the bug. Right now, the only reason this test case passes is because of the code you see in the image below (part of this PR). Once dev fixes the bug in the UI, then I will uncomment that line of code and this test case will prove that the bug is fixed and does not regress.

image

cypress/tests/CreateModels.js
SDESkowronski: See prior answer.

PR #995, opened 2019-02-14
fix: show L format for a previous day LTS for current day

mattmazzola:
The current code keeps showing the `LTS` format if the dialog is less than 24 hours old, but this means you can have time stamps in the list that appear to be in the future which should not be possible.

Example: I create a dialog yesterday at 4:15 PM and I look in the list today and still see 4:15 PM even though it's only 12:50 PM. This is because it still less than 24

This changes is so it only shows `LTS` for dialogs created on the current day. Examples below:

Occurred today: "11:59:24 AM"
Yesterday or before: "02/14/2019"


mattmazzola:
Did you only request changes to indicate @SDESkowronski should see this or what changes did you want?

What was his opinion? This doesn't change how it looks only the conditions it's shown in so if he has code looking specifically for the 'LTS' or 'L' format I would think they would still work.

mattmazzola:
Hmm, I would be interested to see the sites that were presenting times like this. I don't understand how you would understand that an item was yesterday.  If you have no other time related information about the item and you didn't create it or don't remember it would look like: A is `4:00 PM` and B is `12:30 PM` even though A is yesterday and B is actually after / later than A.

As for tests that seems to restrictive.  I would imagine there is a test for times like "Train Dialogs created less than or equal to 24 hours ago show `L` format" and "Train dialogs created more than 24 hours ago show `LTS` format".  All the other tests shouldn't care about the format only that the time is correct since the above test already cover that.  I would say just test for both for both formats. I assume it's getting the text content and trying parse it back into a time.  Can try L and if it doesn't match try LTS and if it still doesn't match fail.

PR #997, opened 2019-02-15
fix: update all reducers to use immer

mattmazzola:
From website:
> You can interact with your data by simply modifying it while keeping all the benefits of immutable data.

> Deep updates in the state of React components can be greatly simplified as well by using immer.

Basically means we can eliminate need to do all the spread `...` stuff to simulate what we want to happen and can simplify our reducers.

Most of the complexity was in the `appReducer` that had to find the app, make a new app, then replace it in the list of apps.  Now it's just find the app and update it.

I left the `replace` and `filter` stuff the way it is since it's more risky to change although that could be updated too.

See:
https://github.com/mweststrate/immer#returning-data-from-producers
https://github.com/mweststrate/immer#example-patterns

Basically anywhere we had to do updated of nested objects this would help. Haven't look much at other cases like `setState` but will save those for another PR anyways.


src/reducers/appsReducer.ts
```diff
@@ -8,98 +8,94 @@ import { Reducer } from 'redux'
...
-
-            const newApp: App = {
-                ...(app as App),
-                lastModifiedDateTime: `${new Date().toISOString().slice(0, 19)}+00:00`,
+                return
             }
 
-            return { ...state, all: replace(state.all, newApp, a => a.appId) }
-
+            app.lastModifiedDateTime = `${new Date().toISOString().slice(0, 19)}+00:00`
```
pnwbrian: new functionality

mattmazzola: What's new?

Previously it was assigning to the `newApp` object. Now it assigns to the original `app` but it was always updating the time.

Btws, why do we force the offset to be 0 like `.slice(0, 19)}+00:00`?
To me it seems misleading as that's not actually when the time was updated.

mattmazzola: Responding again to force notification

PR #998, opened 2019-02-15
fix: use async await where possible

mattmazzola:
Not required but thought it is improvement to keep up with advances in typescript. Previously they didn't allow async/await when targeting es5 but since 2.1 they do.

I think biggest benefit was standardized error handling. You can use same try/catch pattern for synchronous and asynchronous code

See 2.1 release annoucement: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html

VSCode can suggest it like this:
image

For the catch blocks I follow `const error = e as Error` to get type information as recommended here:
https://github.com/Microsoft/TypeScript/issues/8677


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -68,7 +68,7 @@ const convertEntityIdsToTags = (ids: string[], entities: CLM.EntityBase[]): OF.I
 }
 
 // Entities that can be chosen for required / blocking
-const conditionalEntityTags = (entities: CLM.EntityBase[]): IConditionalTag[] =>  {
+const conditionalEntityTags = (entities: CLM.EntityBase[]): IConditionalTag[] => {
```
mattmazzola: Hmm, don't know why this formatting change is showing up here. I thought I did a pass to format all the files to prevent this.  I must have missed some or maybe these got changed in later comments with the new entity stuff.

pnwbrian: I'm surprised as well given VSC autoformatting should have set this correct the first go 'round

src/components/modals/AppCreator.tsx
```diff
@@ -44,20 +44,18 @@ class AppCreator extends React.Component<Props, ComponentState> {
...
 
-    componentDidMount() {
-        getLuisApplicationCultures()
-            .then(cultures => {
-                const cultureOptions = cultures.map<OF.IDropdownOption>(c =>
-                    ({
-                        key: c.cultureCode,
-                        text: c.cultureCode,
-                    }))
+    async componentDidMount() {
```
pnwbrian: I believe the change will introduce a second, extra rendering of the DOM that executes after the browser updates. I remember noting this in the ReactJS docs

mattmazzola: Hmm, I'm not understanding what would be different then the original. Can you explain more about which part of the change makes this behave different and what would you mean by browser update?

I think it would have always been two renders. One initially when we mount the component and another when the `cultureOptions` are loaded and `this.setState` is called.  I think any time props or state is updated `render` is called.  If the render output is the same as before there is no effect / work done by browser.  It was async in the old code and async in the new code we're just using nicer syntax now. 

Only issues I see from this change is methods new return promises instead of undefined but nothing should be depending on the return value so it shouldn't matter.

mattmazzola: I tested and seems to behave as expected. I would be interested to read more about it on the the docs you found though.

As far as I can tell the extra render comes from parent component loading the applications and the timings just happen to line up that it's finishes after the culture request

See: 
image

PR #1000, opened 2019-02-18
Double end session

SDESkowronski:


mattmazzola:
How does this affect the double end session issue?

I see two new calls to: `VerifyScoreActionsButtonIsMissing` and `VerifyTypeYourMessageIsMissing` but I'm not sure how this would prevent having two end sessions in the same train dialog.  

I didn't think Brian has merged any fix for this yet since were waiting to see what Lars thought.  I wanted to show it in the actions list as disabled, but Brian wanted to prevent you from creating it.

SDESkowronski:
This is not a complete fix for this test case, however I wanted to put the changes in since I had done the investigative work to see how a user could potentially screw it up...or how a regression in the UI could allow it to happen. More to come once the bugs are addressed.

This verifies that the user cannot add another session using one of two possible ways...
1. Was the ScoreActions button on a non-wait EndSession action to a training that was saved. Since I've seen this come up, I want to confirm it does not regress and come up in the future. Currently I can't make it happen due to some change in the UI since I saw that. And that is good since the test will pass.
2. Is via the usual chat input as in the image below.
image

mattmazzola:
Ok, well in the future try to make the title and description of the PR describe what the PR does.

For example:
Title: fix: verifies user cannot add more than one end session
Description: Add two checks to... (your 1. and 2. above)

"Double end session" is really generic so you would expect a PR that deal with every possible double end session scenario. Since the expectations of readers are looking for this but all they see is a couple checks added it causes the question.

We are privileged on our team to be able to merge our own branches but this is often not the case. As the name, Pull Request, implies you are Requesting a contributor to Pull in your code changes into their branch.  It helps to get into this mindset since it makes you think, "If I wanted to get my code approved I would want to take the least amount of effort as possible for them to review it and will provide enough relevant information". Of course you have to weigh effort / time spent in crafting the PR vs the actual code change and balance that. It also optimizes the team. If I submit a poor quality PR I know it's going to affect at least 2 or 3 people's time trying to reading and figure out what happened so for sake of optimizing the team it's in my best interest to write my PR's well

PR #1004, opened 2019-02-20
fix: inner text function behavior on electron browser

SDESkowronski:
Fixes a bug that Matt noticed on CircleCI which is using the Electron browser. The element.innerText property contained New Line characters in them in certain situations.

### DIFF
--- a/cypress/support/Helpers.js
+++ b/cypress/support/Helpers.js
@@ -26,7 +26,7 @@ export function StringArrayFromElementText(selector, retainMarkup = false) {
   ConLog(`StringArrayFromElementText(${selector})`, elements.length)
   let returnValues = []
   for (let i = 0; i < elements.length; i++)  {
-    let text = retainMarkup ? elements[i].innerHTML : elements[i].innerText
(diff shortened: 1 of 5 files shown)


cypress/support/Helpers.js
```diff
@@ -26,7 +26,7 @@ export function StringArrayFromElementText(selector, retainMarkup = false) {
   ConLog(`StringArrayFromElementText(${selector})`, elements.length)
   let returnValues = []
   for (let i = 0; i < elements.length; i++)  {
-    let text = retainMarkup ? elements[i].innerHTML : elements[i].innerText
```
mattmazzola: Been meaning to ask why do we use `innerText` instead of `textContent`?

> As a result, innerText is much more performance-heavy: it requires layout information to return the result.

> innerText will turn <br> elements into newline characters, while textContent will just ignore them.

In most cases it seems like we don't care about the layout and only care that the content in the node is the correct content.  I would think this could improve the speed and remove the need for code like this.

SDESkowronski: Because I was unaware of `textContent`...so thanks for pointing that out.

PR #1006, opened 2019-02-21
Action Unavailable Error Reporting

SDESkowronski:
Verifies that we get appropriate error reporting when a Train Dialog is edited to remove a required entity and thus make it unavailable.


SDESkowronski:
You are right, that particular change was minor so I did not isolate the change like I should have. Even though it was minor you went to great lengths to try and figure it out and I could have saved you that hassle. Sorry, I'll do better next time.

PR #1007, opened 2019-02-21
Missing Action Error Handling

SDESkowronski:
Tests error handling for Missing Action


cypress/support/Actions.js
```diff
@@ -27,6 +26,13 @@ export function CreateNewAction({ response, expectedEntities, requiredEntities,
   if (requiredEntities) actionModal.TypeRequiredEntities(requiredEntities)
   if (disqualifyingEntities) actionModal.TypeDisqualifyingEntities(disqualifyingEntities)
   actionModal.ClickCreateButton()
+}
+
+export function CreateNewActionThenVerifyInGrid({ response, expectedEntities, requiredEntities, disqualifyingEntities, uncheckWaitForResponse, type = 'TEXT' }) {
```
mattmazzola: Maybe not required for this review but is related to the idea that the functions/helpers should be further decoupled. In other words, you should be able to control when you create/add an action independently of when you verify it's representation in the list.

I think previously `CreateNewAction` was implicitly doing both things.  Now you made it so you can call `CreateNewAction` without verifying.  Likewise, you would also want to be able to verify an action in the grid with being forced to create one.

Example:
```
const newAction = { ... }
actions.CreateNewAction(newAction )
actions.VerifyInGrid(newAction)
```
Still ensures same is created and verified.

This is an example of something that would be more in line with our plan to migrate the tests to a more flexible feature based organization.  If we change how actions are displayed and modify the list or remove it but don't touch anything related to creating actions the function for creating actions should still work.

cypress/tests/ErrorHandling.js
```diff
@@ -178,3 +181,44 @@ export function ActionUnavailableErrorHandling()
...
+  train.Save()
+
+  modelPage.NavigateToActions()
+  actions.DeleteAction(common.whatsYourName)
+  modelPage.NavigateToTrainDialogs()
+
+  modelPage.VerifyErrorIconForTrainDialogs()
+  train.VerifyErrorsFoundInTraining(`${String.fromCharCode(59412)}${gonnaDeleteAnAction}`, gonnaDeleteAnAction, undefined)
+
+  train.EditTraining(`${String.fromCharCode(59412)}${gonnaDeleteAnAction}`, gonnaDeleteAnAction, undefined)
```
mattmazzola: Ideally we should find a way to remove the need to do this, but since it's a work around you should at least add a constant to provide semantic meaning to why the code is there.

At the top of this file, or in the common file do something like this:
```
const invalidDialogIcon = String.fromCharCode(59412)
...
train.EditTraining(`${invalidDialogIcon}${gonnaDeleteAnAction}`, gonnaDeleteAnAction, undefined)
```

Now people reading this line in the test what this string is added for because they know what it represents.

pnwbrian: I would second the use of a constant here

cypress/support/Actions.js
SDESkowronski: Lets discuss this point via email...and later.

cypress/tests/ErrorHandling.js
SDESkowronski: Good point...should have thought of that myself. I'll fix it.

PR #1011, opened 2019-02-23
fix: ensure slate placeholder cursor is centered

mattmazzola:
The correct fix is for Slate to update and fix the bug, but in the mean time this appears to work.

See: https://github.com/ianstormtaylor/slate/issues/1436

Action Creator Editor on TEXT:

Before:
image

After:
image

on END_SESSION
Before:
image

After:
image

### DIFF
--- a/src/routes/App.css
+++ b/src/routes/App.css
@@ -198,4 +198,8 @@
     color: var(--cl-color-error);
 }
 
+[data-slate-editor="true"] span[contenteditable="false"] {
+    float: left
+}
+
 


mattmazzola:
It's unrelated to the fix but I was curios why we don't put a placeholder in the "Data" field of the END_SESSION action.

It seems like it was intentional:
image

mattmazzola:
@LarsLiden What was your opinion on the placeholder?
Do you remember why it is set to empty string?  I was thinking it was just mistake and since we didn't now what to put at the time we left it blank or maybe it was done intentionally and wanted to know why

LarsLiden:
Unfortunately I don’t remember.  You could try removing it but you’d have to test a bunch of scenarios to make sure nothing breaks.

PR #1012, opened 2019-02-25
1707 block duplicate actions

pnwbrian:
Reject duplicate actions by comparing the pending action map against a JSON stringified versions of the existing action set. Comparison leverages the fast-json-stable-stringify package


pnwbrian:
Unit tests inbound through a second PR

src/Utils/util.ts
```diff
@@ -93,4 +94,10 @@ export function earlierDateOrTimeToday(timestamp: string): string {
     const dialogTime = moment(timestamp)
     const isDialogCreatedToday = dialogTime.diff(endOfYesterday) >= 0
     return dialogTime.format(isDialogCreatedToday ? 'LTS' : 'L')
+}
+
+export function isNewActionUnique(new_action: CLM.ActionBase, actions: CLM.ActionBase[]): boolean {
```
mattmazzola: camelCase

src/Utils/util.ts
```diff
@@ -93,4 +94,10 @@ export function earlierDateOrTimeToday(timestamp: string): string {
...
+    const needle = stringify({ ...new_action, actionId: "", createdDateTime: "" })
+    const haystack = actions.map(action => { return stringify({ ...action, actionId: "", createdDateTime: "", suggestedEntity: action.suggestedEntity || null, packageCreationId: 0, packageDeletionId: 0, version: 0 }) })
```
mattmazzola: Put the properties on separate lines for better readability:

Example:
```
const haystack = actions.map(action => stringify({
    ...action,
    actionId: "",
    createdDateTime: "",
    suggestedEntity: action.suggestedEntity || null,
    packageCreationId: 0,
    packageDeletionId: 0,
    version: 0 }))
```

src/Utils/util.ts
```diff
@@ -93,4 +94,10 @@ export function earlierDateOrTimeToday(timestamp: string): string {
     const dialogTime = moment(timestamp)
     const isDialogCreatedToday = dialogTime.diff(endOfYesterday) >= 0
     return dialogTime.format(isDialogCreatedToday ? 'LTS' : 'L')
+}
+
+export function isNewActionUnique(new_action: CLM.ActionBase, actions: CLM.ActionBase[]): boolean {
+    const needle = stringify({ ...new_action, actionId: "", createdDateTime: "" })
```
mattmazzola: Optional, but instead of forcing the properties you want to ignore to be the same (empty string) on newAction and other actions I would think it makes more sense to remove them.  Seems to better convey intent of NOT comparing these properties since they are expected to be different.

src/react-intl-messages.ts
```diff
@@ -533,6 +535,8 @@ export default {
         [FM.ACTIONCREATOREDITOR_WARNING_PAYLOAD]: 'Bot Response is Required',
         [FM.ACTIONCREATOREDITOR_WARNING_ENTITY]: 'Removed reference to undefined Entity',
         [FM.ACTIONCREATOREDITOR_WARNING_NONEMPTYFIELD]: `Non-wait Actions can't have Expected Entities`,
+        [FM.ACTIONCREATOREDITOR_WARNING_DUPLICATEACTION]: `Actions must be unique`,
```
mattmazzola: Hmm, I think some of our error messages are worded this way and I prefer this "[noun] must be [condition]" formate, but since it's shown to the user I think we should try to make it be more actionable and friendly.
I would prefer something like this:

"The action you attempted to create has the same values as another action that already exists on your model. All actions must be unique. Please change the properties or close."

Can probably word it better but point is it tell them what they can do.  Either change the a property or close and discard the action.

LarsLiden:
Please hold off on your PR until unit tests are included.  That way you know the code you are checking in is working.  😊

You can still check into your local branch

src/Utils/util.ts
pnwbrian: new_action is now newAction

src/Utils/util.ts
pnwbrian: Good suggestion

src/Utils/util.ts
LarsLiden: good suggestion.

package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@conversationlearner/models": "0.197.0",
     "@conversationlearner/webchat": "0.145.0",
+    "@types/fast-json-stable-stringify": "^2.0.0",
```
mattmazzola: There should be a corresponding change in the `package-lock.json`.  You builds seem to be failure because of this.

src/react-intl-messages.ts
pnwbrian: Reworded for a more positive tone

package.json
pnwbrian: Crossed paths in the PR. Now listed in the package.json dependency list

mattmazzola: Hmm. I'm still not seeing the associated change in the `package-lock.json`.

`@types/*` dependencies should be `devDependencies`

Also it looks like you are using `fast-json-stable-stringify` so it should also be a dependency.

src/Utils/util.test.ts
```diff
@@ -20,4 +21,201 @@ describe('util', () => {
...
+                requiredConditions: [],
+                requiredEntities: [],
+                requiredEntitiesFromPayload: [],
+                suggestedEntity: null,
+                version: 0
+            }
+
+            let actionSet: CLM.ActionBase[] = []
+
+            // as first action created
```
mattmazzola: These lines looks like descriptions of a test so just make them tests.

src/Utils/util.test.ts
```diff
@@ -20,4 +21,201 @@ describe('util', () => {
...
+            expect(isNewActionUnique(sampleAction, actionSet)).toEqual(true)
+
+            // as absolutely identical action
+            actionSet.push(sampleAction)
+            expect(isNewActionUnique(sampleAction, actionSet)).not.toBe(true)
```
mattmazzola: Use positive assertions when possible.  We want to know exactly what it should be not what it isn't.  It just so happens that it's a boolean and we know it can only be false but don't rely on this.

src/Utils/util.ts
```diff
@@ -5,6 +5,7 @@
 import * as CLM from '@conversationlearner/models'
 import * as IntlMessages from '../react-intl-messages'
 import * as moment from 'moment'
+import * as stringify from 'fast-json-stable-stringify'
```
mattmazzola: As mentioned earlier this needs to be declared as dependency.  

I think you had mentioned some other package already required. We can't depend on this. Always declare what you need, if that package changes and removes their dependency our app would break.

src/Utils/util.ts
```diff
@@ -93,4 +94,15 @@ export function earlierDateOrTimeToday(timestamp: string): string {
     const dialogTime = moment(timestamp)
     const isDialogCreatedToday = dialogTime.diff(endOfYesterday) >= 0
     return dialogTime.format(isDialogCreatedToday ? 'LTS' : 'L')
+}
+
+export function isNewActionUnique(newAction: CLM.ActionBase, actions: CLM.ActionBase[]): boolean {
+    const needle = normalizeActionAndStringify(newAction)
+    const haystack = actions.map(action => normalizeActionAndStringify(action))
+    return haystack.findIndex(straw => straw === needle) < 0
```
mattmazzola: use `some` instead of `findIndex`.  You're not looking for an index your trying to see if anything is equal

src/Utils/util.ts
```diff
@@ -93,4 +94,15 @@ export function earlierDateOrTimeToday(timestamp: string): string {
     const dialogTime = moment(timestamp)
     const isDialogCreatedToday = dialogTime.diff(endOfYesterday) >= 0
     return dialogTime.format(isDialogCreatedToday ? 'LTS' : 'L')
+}
+
+export function isNewActionUnique(newAction: CLM.ActionBase, actions: CLM.ActionBase[]): boolean {
```
mattmazzola: Hmm, questioning putting this in utils given the context. If you're going to put this as a utility it needs to be as re-usable and generic as possible.

Would drop the "new" and rename as `isActionUnique`.  You don't know where this method will be used and can't guarantee it will be for "new" actions.  For example, we should probably also protect against any modification of an action that would make it not unique so in that case it would be `isEditedActionUnique` which doesn't make sense.

src/Utils/util.ts
```diff
@@ -93,4 +94,15 @@ export function earlierDateOrTimeToday(timestamp: string): string {
...
+    const needle = normalizeActionAndStringify(newAction)
+    const haystack = actions.map(action => normalizeActionAndStringify(action))
+    return haystack.findIndex(straw => straw === needle) < 0
+}
+
+function normalizeActionAndStringify(newAction: CLM.ActionBase) {
```
mattmazzola: Hmm, not a fan of the word normalize but I could see it.  Will see what others say.

Options, but I would prefer something like `removeIncomparableProperties`

Also would keep the functions split. You generally wnt things to be composable and to small things well.  I may want a stripped action but not the string version of it.  E.g. `removeIncomparableProperties(stringify(action))`

src/react-intl-messages.ts
mattmazzola: I see you changed `FM.ACTIONCREATOREDITOR_WARNING_DUPLICATEFOUND`, but left `FM.ACTIONCREATOREDITOR_WARNING_DUPLICATEACTION`.  Why?

It looks like the `FM.ACTIONCREATOREDITOR_WARNING_DUPLICATEACTION` is showing to the user in the Confirmation modal up top: https://github.com/Microsoft/ConversationLearner-UI/pull/1012/files#diff-d106c181edc321d094ccd9d2a2c34667R1421

src/Utils/util.ts
pnwbrian: Agree with the more generic name would be IsActionUnique.

src/Utils/util.ts
pnwbrian: Good suggestion as some() is the intent. I'd considered the index at a later date should the UI offer to flag the matching action

src/Utils/util.ts
pnwbrian: I'm inclined to leave this one as coded. The composable and small philosophy is a worthy point going forward, especially when writing new code to the code base.

mattmazzola: Ok.  I didn't notice before, but rename `newAction` parameter to just `action`. This function is used on the set of existing actions as well.

src/Utils/util.ts
mattmazzola: Hmm maybe my comments are not behaving the way I had intended. This was marked as resolved but I'm still seeing the old code.  Also `camelCase`

src/Utils/util.test.ts
pnwbrian: Adds value by documenting the intent of the asserted tests

src/Utils/util.test.ts
pnwbrian: Switching to positive asserts

src/Utils/util.test.ts
```diff
@@ -70,4 +71,221 @@ describe('util', () => {
             expect(actual).toBe(true)
         })
     })
+
+    describe('isActionUniqueTEXT', () => {
+        it('exercise TEXT action creation', () => {
```
mattmazzola: This looks like the only `it` block in the `describe`

I also didn't know it would allow to put `test` wthin an `it`...  

Your tests seems to be structured in a weird way. Try it like this:

```
describe('isActionUnique', () => {
  describe('TEXT actions', () => {
    test('as first action created', () => { ... })
  })

  describe(`ENDSESSION actions`, () => { ... })
  describe(`CARD actions`, () => { ... })
})
```

This poses other question. It seems like we're doing the same tests for different types of actions but the `isActionUnique` doesn't do anything different based on the type. What do these tests cover that the previous didn't?

pnwbrian: Realigned the test to use the nested describes, and lose the it wrappers.

Speaking to isActionUnique itself, the code stringifies most of the action object, including the actionType property, which varies between action types. The product currently has a limited trap on adding duplicate ENDSESSION actions for example. Now, using this code, the duplicate action detection is significantly stronger.

src/Utils/util.test.ts
```diff
@@ -70,4 +71,221 @@ describe('util', () => {
...
+                requiredConditions: [],
+                requiredEntities: [],
+                requiredEntitiesFromPayload: [],
+                suggestedEntity: null,
+                version: 0
+            }
+
+            let actionSet: CLM.ActionBase[] = []
+
+            test('as first action created', () => {
```
mattmazzola: Change these to be the similar language/syntax as the existing tests. They should be something like "given X return Y" or "when X should do Y"

Something similar to:
"given action not in list of actions return true"
"given action when identical action is in list return false"
...

src/Utils/util.test.ts
```diff
@@ -70,4 +71,221 @@ describe('util', () => {
...
+                expect(isActionUnique(sampleAction, actionSet)).toBe(false)
+            })
+
+            test('as a similar, but not identical into very similar set', () => {
+                actionSet = [{ ...sampleAction }]
+                let similar = { ...sampleAction, isTerminal: !sampleAction.isTerminal }
+                expect(isActionUnique(similar, actionSet)).toBe(true)
+            })
+
+            test('as a similar, but not identical into very similar set. how smart is the find operator?', () => {
```
mattmazzola: "how smart is the find operator?"
I didn't notice this comment before. I didn't think this relied on any `find` operations but only `some` and comparing strings.  Anyways remove it.

PR #1013, opened 2019-02-25
chore: change model tag to model version

mattmazzola:
Renames model Tag to model Version to avoid future confusion with Tags on Train Dialogs and to align with LUIS.

**Below model name:**
Before:
image

After:
image

**Log dialog column**
Before:
image

After:
image

I named the column "Model Version" instead of "Version" to make sure people new it wasn't a version of the log dialog.

### DIFF
--- a/src/react-intl-messages.ts
+++ b/src/react-intl-messages.ts
@@ -71,6 +71,8 @@ export enum FM {
     APP_HEADER_DOCS = 'App.header.docs',
     APP_HEADER_SUPPORT = 'App.header.support',
 
+    APP_VERSION = 'App.version',
+
     // TrainingStatus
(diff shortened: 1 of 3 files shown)


src/react-intl-messages.ts
```diff
@@ -71,6 +71,8 @@ export enum FM {
     APP_HEADER_DOCS = 'App.header.docs',
     APP_HEADER_SUPPORT = 'App.header.support',
 
+    APP_VERSION = 'App.version',
```
mattmazzola: Technically this is MODEL_VERSION but everything else was named with APP_ prefix so I followed pattern to avoid confusion.  We renamed APP to MODEL a while back.

PR #1014, opened 2019-02-27
Move all tests into integration folder

SDESkowronski:
This is part 1, I still need to organize tests into smaller stages, which will be easy unless a full rewrite is called for (which at this time Lars said no to doing that).

**This was the minimum changes to accomplish the following:**

- Move all test cases back into the `cypress/integration` folder, named with `.spec.js` extension.
- Remove the TestList.js & TestListManager.js

**Lost:** The persistent logging capability - **To Fix Later**

**Retained:**

- Ability to run a single test case
- Ability to run multiple test case from the Test Runner UI


cypress/integration/(Regression).js
```diff
@@ -3,4 +3,32 @@
  * Licensed under the MIT License.
```
mattmazzola: The file above says `--spec "cypress/integration/(Regression).spec.js"` which I thought intended to refer to this file but I'm not seeing the rename.

I was expecting this file to also be renamed with the `.spec.js` to the command above can find it.

cypress/integration/CreateModels/AllEntityTypes.spec.js
```diff
@@ -0,0 +1,20 @@
+/**
+* Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+*/
+
+const models = require('../../support/Models')
+const entities = require('../../support/Entities')
+
+describe('CreateModels', () => {
+  it('All Entity Types', () => {
```
mattmazzola: Can save for different review, but eventually want the `it` should say what the test is asserting.  Think keeping it as simple port is ok for now.

As some background on the naming. I can't remember the exact origins of the syntax but I think it is call Gherkin or "GivenWhenThen" meaning "Given state X when condition Y then expect Z" or something like that.  
Combining this with Mocha tooling and BDD style it's meant to read well to humans. `describe` is literally describing what the children tests are related to or declares the subject of the tests assume.  Then `it` s read into their phrase like "it should do x"

```
describe('myClass', () => {
  describe('doubleThings', () => {
    it('given a string it should return the string concatenated with itself', () => { ... })
    it('given a number should return double the number', () => { ... })
  })
})
```

cypress/integration/(Regression).js
mattmazzola: It even says in your test run `Can't run because no spec files were found.` ...

SDESkowronski: Good catch. An oversight.

cypress/integration/CreateModels/AllEntityTypes.spec.js
SDESkowronski: I understand that. Was intending to go that way moving forward, but not in this change set.

PR #1015, opened 2019-02-27
chore: add empty placeholder for models list

mattmazzola:
During the process of me having to setup my computer again I had come across an empty models list and realized we don't have a placeholder like we do on all the other pages (Actions, Entities, Dialogs...)

It should help the user during the first run experience when they would not have any models.

Before:
image

After:
image

Modeled after other pages like Entities:
image


LarsLiden:
Nice!

mattmazzola:
I assume "Nice" means you approve?

Also, is there a way to prevent your email from expand the whole thread?

You only commented 1 word but the transferred comment is massive.  Makes it hard to read your comment and the comments below or above.

One option, is to click on the "view in Github" button and then comment to avoid the issue.  Just as fast

LarsLiden:
Yes, approve.  Man that does make email reply pretty much useless

PR #1020, opened 2019-02-28
Fix Persistent Logging

SDESkowronski:
Persistent logging was broken when I removed the TestListManager. This code fixes it.

Also this demonstrates how to cause dependent tests to be skipped when a test that runs before it fails. It is easy to do. 

I included a new Tools test case `TestUiViewAndLogs.spec.js` that has multiple test suites and it tests both the features of the Persistent Logging and skipping subsequent tests when one of them fails.

Both of the exploit Mocha data about the test cases.

To see the results run `TestUiViewAndLogs.spec.js` and then look in the folder `results/cypress` for the log files. Below are some images from the Test Runner UI...I like this much better than the prior way since we don't have failures that are not real.

image
image


cypress/integration/Tools/TestUiViewAndLogs.spec.js
```diff
@@ -0,0 +1,84 @@
...
+// -----------------------------------------------------------------------------
+// The Purpose of this Test Spec File is to test the Persistent Logging features
+// and to demonstrate how to skip dependant tests when their predicessor fails.
+// -----------------------------------------------------------------------------
+
+const helpers = require('../../support/Helpers')
+
+// The difference between the 1st and 2nd is where we place the call to SkipRemainingTestsOfSuiteIfFailed()
+describe('1st Test Suite Root', () => {
+  afterEach(helpers.SkipRemainingTestsOfSuiteIfFailed)
```
SDESkowronski: Placing the `Skip...` here causes all remaining tests of the suite to be skipped if there is a failure.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    it('The FIFTH Test', () => {
+      expect(true).to.equal(true)
+    })
+  })
+})
+
+// The difference between the 1st and 2nd is where we place the call to SkipRemainingTestsOfSuiteIfFailed()
+describe('2nd Test Suite Root', () => {
+  describe('Level 2-1', () => {
+    afterEach(helpers.SkipRemainingTestsOfSuiteIfFailed)
```
SDESkowronski: Placing the `Skip...` here causes all remaining tests of the **SUB**suite to be skipped if there is a failure, yet the tests of the level one up will continue to run.

package.json
```diff
@@ -80,7 +80,7 @@
     "@types/uuid": "^3.4.4",
     "commitizen": "2.10.1",
     "cross-env": "^5.2.0",
-    "cypress": "^3.0.3",
+    "cypress": "^3.1.5",
```
SDESkowronski: Oh yes, this requires an upgrade to the latest version of Cypress. Necessary to do a cy.writeFile with the {flag: 'a'} option in it. (appending to the file)

cypress/integration/Tools/CreateModel.spec.js
```diff
@@ -6,8 +6,8 @@
 const models = require('../../support/Models')
 const homePage = require('../../support/components/HomePage')
 
-describe('Tools', () => {
-  it('Create Model', () => {
+describe('Create Model', () => {
+  it('Tools', () => {
```
pnwbrian: I believe Matt would like "it" changed to "test"

mattmazzola: What lead you to believe this?  

The function you use/choose corresponds how the tests are written.  If writing style is more BDD then I think `it` fits better if it's more traditional / unit then `test` might be more appropriate.

cypress/support/PersistentLogs.js
```diff
@@ -16,15 +21,98 @@ console.log = function (message) {
...
+// TRUE `beforeEach` CODE FOR THE TEST CASE THAT IS JUST NOW STARTING
+// This MUST come after the prior `beforeEach` code block and cannot
+// be combined with that other code block due to the `cy.writeFile`
+// call that it contains.
+beforeEach(function() {
+  console.log('')
+  console.log(`******************** ${GetFullTestSuiteTitle(this.currentTest)} ************************************************************`)
+
+  test = this.currentTest
+  let title = GetSuiteRootTitle(this.currentTest)
```
mattmazzola: This function could return arbitrarily long string or strings with spaces. At the very least it wouldn't be a good candidate for filenames.  

On larger point can you explain a bit more about these logs:
What do these capture that Cypress doesn't capture?
Do these log relate: https://docs.cypress.io/api/events/catalog-of-events.html#Logging-All-Events
Is this just temporary until cypress adds the feature to save logs?

cypress/support/PersistentLogs.js
```diff
@@ -16,15 +21,98 @@ console.log = function (message) {
...
+  FlushLogEntries() 
 })
+
+// -----------------------------------------------------------------------------
+// Support functions.
+// -----------------------------------------------------------------------------
+
+function LogTestState()
+{
+  if (!test) { return false }
```
mattmazzola: Looks like it relies on a global variable. Can you try to achieve this without globals?

Seems like anyone could overwrite this to something arbitrary like `test = { foo: "bar" }` and this line would pass and try to use it.

cypress/support/PersistentLogs.js
```diff
@@ -16,15 +21,98 @@ console.log = function (message) {
...
+  console.log("'".repeat(message.length))
+}
+
+// NOTE: This requires the latest version (3.1.5) of Cypress:
+//        change directory to the root project folder
+//        npm install cypress --save-dev
+function WriteLogEntries() { 
+  if (logFileName) {
+    cy.writeFile(logFileName, logEntries, {flag: 'a'})
+      .then(() => { logEntries = '' })
```
mattmazzola: Looks like it uses more globals.  A function should only be using arguments that were passed in or local variables that were declared inside it.

cypress/support/PersistentLogs.js
```diff
@@ -16,15 +21,98 @@ console.log = function (message) {
...
+
+function GetSuiteRootTitle(test) {
+  if (!test.parent) throw new 'Did not find test.parent'
+  
+  if (test.parent.title === '') { return test.title }
+  return GetSuiteRootTitle(test.parent)
+}
+
+function GetFullTestSuiteTitle(test) {
+  if (!test.parent) throw new 'Did not find test.parent'
```
mattmazzola: Seems to repeat the same code as above. Is there a way to combine these?

cypress/integration/Tools/TestUiViewAndLogs.spec.js
mattmazzola: Would need to make the support the case where people want to write other things in the `afterEach`

Like:
```
afterEach(() => {
  someCleanUp()
  otherCustomCode()
  helpers.SkipRemainingTestsOfSuiteIfFailed()
})
```

I imagine you'll have to pass in stuff from the outer `this`.  Also we want to support short hand functions `() => { ... }`.  I think cypress has alternatives using `this`

SDESkowronski: You bring up some interesting points:

1. JavaScript does not give you the "this" object in the same way when you use arrow function as when you use named or unnamed functions. Thus whenever you use the `SkipRemainingTestsOfSuiteIfFailed` function you cannot use arrow functions (lambdas). So you can do what you want if you change to this:

```suggestion
afterEach(function() {
  someCleanUp()
  otherCustomCode()
  helpers.SkipRemainingTestsOfSuiteIfFailed()
})
```
2. Under what circumstances would you need to do cleanup? It is highly recommended to avoid cleanup in the test code since we cannot guarantee that it will happen. And if we depend on it happening in future tests then we have a big mess.

3. You can also repeat all of the hooks since they add them to some sort of event queue. So if necessary you could have one afterEach just as I specified it, and then a second one to deal with your other needs...I found this handy when I had to make sure all of the Cypress `cy` calls were completed before moving on to the next bits of code that needed to be executed after the `cy` functions finished.

SDESkowronski: I did not understand this part of your comment...
> I think cypress has alternatives using this

cypress/support/PersistentLogs.js
SDESkowronski: Yes these logs are temporary...unless Cypress' future implementation does not work for us. Even in that case we can probaly work through it.

Persistent Logs are for the console.log output. Here are some examples:
> 
-+- 00:50:39..744 - cy.Get([data-testid="train-dialogs-turns"]) - Start - Last DOM change was 787 milliseconds ago
-+- 00:50:39..776 - cy.Get([data-testid="train-dialogs-turns"]) - DOM Is Stable
> 
The above log tells me what UI element we are getting and also shows timing of when the DOM changes, so if we somehow miss something I either need to adjust the algorithm (which has not happened in months) or there is some funky UI rendering going on that I need to file a bug on.

> 
-+- 00:50:39..816 - StringArrayFromElementText([data-testid="train-dialogs-first-input"]) - 1
-+- 00:50:39..816 - StringArrayFromElementText([data-testid="train-dialogs-first-input"]) - Hello
> 
These logs are from a part of the code that often fails. Typically when making a mistake calling the functions using this, but also when unexpected UI elements are present like newlines or that error icon. So this shows what exactly was found so that test code can use the appropriate values.

SDESkowronski: As for the length of the root title...

I believe:
1. We should keep these as short as possible. 
2. We should name our spec file with this same title.
3. Even if it is long, the file system can handle it.

I can and should remove the spaces and camel case it. Given my beliefs about this, I think it makes a perfect name for the logs, makes it easy to figure out which log goes with which test case. I've used this logging mechanism on every project I've been on that I have designed the logging on (all 3 of them) and nobody else complained about it. I've seen others use it in a similar way too.

cypress/support/PersistentLogs.js
SDESkowronski: It is only global to the scope of this code file, it is not exported. It is defined as:
> var test
> 
at the beginning of the file.

cypress/support/PersistentLogs.js
SDESkowronski: Same response as above.

cypress/support/PersistentLogs.js
SDESkowronski: In one method I need just the root parent title, this is for the log file name and knowing when I need to start a new log file.

In the other case I need the full chain of titles with a dash separating them for visual readability since they appear in the logs to separate test stages in the log.

I sent you an example of a log file on teams. It will help you understand why I need the code the way it is. I don't think combining the functionality into one function would be nice, it can be done, but I need two different returns from each for different purposes. The function names state the purpose and make the code more readable.

cypress/support/PersistentLogs.js
mattmazzola: Yes, which is a global ... and also uses `var` both things we don't want.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
mattmazzola: > JavaScript does not give you the "this" object in the same way when you use arrow function

Yes, that's why we shouldn't rely on `this` since it's not always that you think.  We don't want to have to remember this special case not being able to use lambdas when write an `afterEach`

> Under what circumstances would you need to do cleanup?

Hmm, i had suspicion that example would confuse you.  Only chose it because that's what most people do in `afterEach`.  The code can be anything.  It's just a function. For example in your code you do logging.

cypress/support/PersistentLogs.js
mattmazzola: > We should keep these as short as possible.

More important to be meaningful than short.

> Even if it is long, the file system can handle it.

Hmm, I thought there were versions of Windows that still the 260 max path limit.  I thought if the developer cloned the repo in a location relatively deep and we try to write a file which adds additional characters we might reach this limit and fail.  I didn't want to risk someone not being able to run the tests simply be cause they cloned repo in certain location or if they wrote a long test name. In any case this will probably be rare.

> I can and should remove the spaces and camel case it

Were you intending to do this? Or leaving it?

I had seen this line:
https://github.com/Microsoft/ConversationLearner-UI/pull/1020/files#diff-e20a789bb1cfa9903bd0f7f355cc850dR62

which does some concatenation with periods for other parts like `YY.MM.DD.HH.mm.ss..SSS` and thought you were doing it to prevent spaces.

If it can't fail then I don't care as much about the name.

cypress/support/PersistentLogs.js
```diff
@@ -16,15 +21,98 @@ console.log = function (message) {
...
+function FlushLogEntries() {   
+  if (logFileName && logEntries.length > 0) {
+    // Because of 'afterEach' functions that were called after we wrote our previous
+    // log entries, there are additional log entries to persist to that same log file.
+    WriteLogEntries()
+  }
+}
+
+function GetSuiteRootTitle(test) {
+  if (!test.parent) throw new 'Did not find test.parent'
```
mattmazzola: Does the code throw new 'Did not find test.parent' actually run without causing it's own error?

I didn't think you could follow the new keyword with a string. new is for creating new objects from constructor functions like: const myInstance = new MyClass() . I don't know what new "asbasdf" does.

In any case, always throw actual Error objects to ensure you get the proper stack trace and the consumers of errors are guaranteed the format. Should also make the error messages as helpful/informative as possible.

E.g. throw new Error(`There is no parent for test: ${test.title}`) tells them which test didn't have a parent to help investigating.

SDESkowronski: The new is a mistake...a throwback to my C# days.

Good point on adding `test.title` to the error, I'll do it.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
SDESkowronski: I looked for an alternative in the Cypress docs to the `this` but found none.

I just added comments to remind them AND an exception if they call it wrongly so as to remind them of the correct way to call it.

> Test Code Error: Cannot use arrow/lambda function to call the SkipRemainingTestsOfSuiteIfFailed() function.
>

cypress/support/PersistentLogs.js
SDESkowronski: Actually it is nice with the spaces in, so I would rather leave it.

As for the periods separating year, month, day, etc...I like having them as separators as it helps make sense of what you are looking at when trying to distinguish one log from another with the same root name.

I spent about an hour trying to make it not fail if someone has a path that is too long, could not do it. But I really don't think we should just throw this functionality out if someone makes their root path too long and without writing the logs to a file this entire code file is useless.

In my experience with test development, most developers understand this restriction and don't do that, and those that do spend a few hours getting a lesson in what not to do that prevents them from repeating that mistake in the future. Such is the life and learnings of a developer.

cypress/support/PersistentLogs.js
SDESkowronski: I wrapped it all in a nameless function scope and removed the var/globals.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
```diff
@@ -0,0 +1,84 @@
...
+  describe('Level 2-1', () => {
+    it('The FIRST Test', () => {
+      expect(true).to.equal(true)
+    })
+
+    it('The SECOND Test', () => {
+      expect(false).to.equal(true) // Intentionally fail to verify that the next tests will be skipped.
+    })
+
+    it('The THIRD Test', () => {
```
mattmazzola: Related to our discussion about declaring dependencies.  How would a developer declare this test ('The THIRD Test') to not have dependencies on "The FIRST Test" or "The SECOND Test"?

Say the `before` / `beforeEach` are successful and the test is setup correct. Test 2 depends on Test 1. Test 1 fails so Test 2 does not run, but Test 3 was NOT dependent so Test 3 still runs.

As it looks now it would skip all tests after the failed test.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
mattmazzola: So did you end up making this work when it is mixed with other functions? E.g. more than skipping tests in the `afterEach`.

Otherwise "can't use lambdas" is only 1 of the special cases and the other is "you also can't do anything else in this function".

SDESkowronski: The answer is yes, it does work when mixed with other function calls. I had already addressed this in my first reply #1. Not only that but you can also supply multiple calls to `afterEach` with each one containing different functionality, both will be executed in the order that `afterEach` was called.

cypress/integration/Tools/TestUiViewAndLogs.spec.js
SDESkowronski: To do that the 3rd test would have to be in a `describe` scope that the 1st and 2nd test is described in and the `SkipRemainingTestsOfSuiteIfFailed` is defined within. So yes this is easily doable. An example of this is in the `2nd Test Suite Root` where the 4th and 5th tests do continue to run even if there is a failure in the 1st, 2nd or 3rd test.

However, in all of the tests I have written so far, all of the subsequent test fragments do depend on the prior tests passing in order to carry on. I doubt there will be many cases that we can come up with where this dependency does not exist.

SDESkowronski:
I included the `SkipRemainingTestsOfSuiteIfFailed` because I discovered the possiblity of doing it while researching how to get data about the test being run from Mocha so I could do the logging changes...AND because it was a simple 1 helper function addition along with adding 2 examples of usage to code I already included to verify the new logging code. 

Also to economise on code review time, especially the lag time between when you have the time to do the review and I have the time to respond. However, if you'd rather I make even simple changes such as this an individual code review then I can do that in the future.

PR #1021, opened 2019-03-01
feat: final enum implementation

LarsLiden:


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -67,6 +67,37 @@ const convertEntityIdsToTags = (ids: string[], entities: CLM.EntityBase[]): OF.I
         .map<OF.ITag>(convertEntityToTag)
 }
 
+const toConditionName = (entity: CLM.EntityBase, enumValue: CLM.EnumValue): string => {
+    return `${entity.entityName} = ${enumValue.enumValue}`
```
mattmazzola: Seems really weird to have the save variable twice.  I'm guessing the parent isn't actually an enumValue it's the container of the value?

I thought it was like:
```
{
  id: "adfadf"
  value: "asdfasdf"
}
```

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -323,7 +355,7 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                 const action = nextProps.action
 
                 const payloadOptions = this.props.entities.map(convertEntityToOption)
-                const negativeEntityTags = convertEntityIdsToTags(action.negativeEntities, nextProps.entities)
+                let negativeEntityTags = convertEntityIdsToTags(action.negativeEntities, nextProps.entities)
```
mattmazzola: Why is this changed to `let`?  

Everything should be `const` unless you intend to reassign it later do to modification.

When I see `let` I think "ok this variable is required to be mutated below" then I have to remember it for the rest remainder of reading.  I see the `push` but not a re-assignment.  If that was only case, you can still push into a const and maintain assurance that no one will change the variable entirely like `negativeEntityTags = []`

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -178,12 +181,22 @@ class Container extends React.Component<Props, ComponentState> {
...
                 })
             }
         }
     }
 
+    initEnumValues(enumValues: CLM.EnumValue[] | undefined): (CLM.EnumValue | null)[] {
+        if (!enumValues) {
+            return Array(CLM.MAX_ENUM_VALUES).fill(null)
+        }
+        const enumClone: CLM.EnumValue[] = JSON.parse(JSON.stringify(enumValues))
```
mattmazzola: Something weird is going on here. 

On the top the type is:
`enumValues: CLM.EnumValue[] | undefined`
then you clone it an caste to:
`enumClone: CLM.EnumValue[]`
but then later have to check that it's undefined
`...enumClone || []`

Seems like the middle cast should preserve what the input was `CLM.EnumValue[] | undefined` because if `enumClone` was truly an array it could never be `null` or `undefined` to trigger the `||`

mattmazzola: Why is the clone required?

Is this so they can edit the values in the view without updating the actual original values?

We should try to limit the use of `JSON.parse(JSON.stringify(`.  It's a rather expensive hack to clone things in javascript so we want to only use it when needed.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -28,13 +28,14 @@ const initState: ComponentState = {
     isMultivalueVal: false,
     isNegatableVal: false,
     isEditing: false,
-    enumValues: [undefined, undefined, undefined, undefined, undefined],
```
mattmazzola: Glad to see this gone. :)

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -203,6 +222,18 @@ class Container extends React.Component<Props, ComponentState> {
...
+    areEnumsIdentical(newEnums: CLM.EnumValue[], oldEnums: CLM.EnumValue[]): boolean {
+        // If any new enums, or old ones changed or deleted
+        if (newEnums.every(ev => ev.enumValueId !== undefined) &&
+            oldEnums.every(oldEnum => {
+            let newEnum = newEnums.find(ne => ne.enumValueId === oldEnum.enumValueId)
+            return (newEnum !== undefined && newEnum.enumValue === oldEnum.enumValue)
+        })) {
+            return true
+        }
+        return false
```
mattmazzola: Do we need the outer `if`?

`every` already returns `true` or `false` and the `&&` ensures both so the outer `if` seems unnecessary.

LarsLiden:
Needed to change it to let as it is now changed in the code later on in the function call.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -337,9 +363,53 @@ class Container extends React.Component<Props, ComponentState> {
...
+                deleteEnumCheck: enumValue
+            })
+            return
+        }
+        else {
+            this.deleteEnum(enumValue)
+        }
+    }
+
+    deleteEnum(enumValue: CLM.EnumValue): void {
```
mattmazzola: Looks like it can be `private`. Ensure people dont' mistakenly call it instead of `onDeleteEnum` since they look similar

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -337,9 +363,53 @@ class Container extends React.Component<Props, ComponentState> {
...
+        let index = this.state.enumValues.indexOf(enumValue)
```
mattmazzola: All these local variables can be `const`. They don't look to be modified below

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -337,9 +363,53 @@ class Container extends React.Component<Props, ComponentState> {
...
+        if (index >= 0) {
```
mattmazzola: I think this method should throw an `Error` if the enum wasn't found.  Don't think we're expecting `deleteEnum` to silently fail in any cases we use it so we can safely enforce that people don't do this by accident.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -337,9 +363,53 @@ class Container extends React.Component<Props, ComponentState> {
...
+        let enumValue = enumValues[index]
+
+        if (newValue.length > 0) {    
+            // Create new EnumValue if needed 
+            if (!enumValue) {
+                enumValues[index] = { enumValue: newValue }
+            }
+            // Otherwise set
+            else {
+                enumValue.enumValue = newValue
```
mattmazzola: Hmm this again.  `enumValue.enumValue`  

I think I understand that the first one is referring to the object and the second is the string? 
If so we should try to call them that, especially when they're mixed so people don't get confused. Like `enumValueObject.value`?

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -502,6 +597,23 @@ class Container extends React.Component<Props, ComponentState> {
...
+    @OF.autobind
+    onCancelEnumDelete() {
+        this.setState({
+            deleteEnumCheck: null
+        })
+    }
+
+    @OF.autobind
+    onConfirmEnumDelete() {
+        if (this.state.deleteEnumCheck) {
```
mattmazzola: Why do we need the `if` statement?

I thought the design was such that this window to confirm the deletion of an enum in use should only be open when there is an enum assigned to.  

If this is true, you could should never be able to click confirm when `deleteEnumCheck` is null.  
If this is false and there is a case where they can be out of sync, the user would not be able to close the window. Since the if statement would skip the logic in here and clicking confirm would not do anything.

src/components/modals/ActionCreatorEditor.tsx
mattmazzola: Somehow your comment got put in different spot so copying it up here for context:

> Needed to change it to let as it is now changed in the code later on in the function call.

Hmm, can you link to the line? I can't find it, only see the `push`

src/components/modals/ActionCreatorEditor.tsx
LarsLiden: Yeah, it's an EnumValue object.   Shahin asked to the name the value "enumValue" too, so it's enumValue.enumValue

src/components/modals/ActionCreatorEditor.tsx
LarsLiden: Yes, the push.  Can't push onto a const, correct?

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: Yeah - need a deep copy.  Spread just copies the object pointers.

LarsLiden: Good catch.  Left over from refactor.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: good catch

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: Done

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: done.

src/components/modals/ActionCreatorEditor.tsx
mattmazzola: You can.  `const` only means single assignment.  You can still mutate the nested properties of objects or array values.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: added console error

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: good suggestion

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
LarsLiden: Fixed so can still close the window

src/components/modals/ActionCreatorEditor.tsx
LarsLiden: ah... thx

PR #1022, opened 2019-03-01
feat: enum feature code  cleanup

LarsLiden:


package.json
```diff
@@ -80,7 +80,7 @@
     "@types/uuid": "^3.4.4",
     "commitizen": "2.10.1",
     "cross-env": "^5.2.0",
-    "cypress": "^3.0.3",
+    "cypress": "^3.1.5",
```
mattmazzola: Hmm, why does this need to be updated?
I understand we want to also use the latest stuff and it probably wouldn't hurt to update but I would think this is not required for enum to function properly.

It also looks like part of another change here: https://github.com/Microsoft/ConversationLearner-UI/pull/1020/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R83

LarsLiden:
Change game from Michael.  Just showed up here because of updating PR after a merge.

package.json
mattmazzola: > Change game from Michael. Just showed up here because of updating PR after a merge.

Yea, so it can be removed? That PR isn't merged yet so I how did it get in here?

The current `master` still has 3.0.3
https://github.com/Microsoft/ConversationLearner-UI/blob/master/package.json#L83

LarsLiden: Michael said in other PR that it's needed

mattmazzola: Yes, for his PR its needed which is why he made the change, but why is it needed for this PR?

It will probably cause conflicts for his PR now since `package-lock.json` is changed.

PR #1025, opened 2019-03-01
fix: move validity indicator icon back to first column

mattmazzola:
Originally the icon was on the first column which was first input, then first input was changed to tags, then tags was moved to be second column instead of first and I forgot to move the triangle back.

This moves it back to the first column like below:
image

Also added a new attribute:
`data-testid="train-dialogs-validity-indicator"` to help make it easier on tests. Previously they were testing for firstInput and having to know if it was valid or not, but the icon isn't actually part of the input so this was misleading.  Even though the first inputs are still in the DOM they don't including the icon so I think there will need to be new way to test this.

Couple questions for further improvement or future PRs:

- **Dates color**
I noticed that we only colorize the scenario, tags, and turns, but not the dates.  My first instinct thought the whole row should be the same color but maybe it was left this way to make reading the dates easier?
Anyways, want to decision to changes those also or leave them.

- **Tag color**
Current it just uses the same red text coloring as the other fields.  An alternate design might be to preserve text contrast and change tag background like below:
image
I thought this was too harsh, but so I left it out but wanted to get other's opinions. If we do it I think it needs to be the consistent red error color everywhere though so just making it a faded red didn't seem like an option.

### DIFF
--- a/src/routes/Apps/App/TrainDialogs.tsx
+++ b/src/routes/Apps/App/TrainDialogs.tsx
@@ -142,9 +142,9 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
             maxWidth: equalizeColumnWidth,
             isResizable: true,
             render: trainDialog => {
-                return <span className={textClassName(trainDialog)} data-testid="train-dialogs-tags">
+                return <span className={`${OF.FontClassNames.mediumPlus}`} data-testid="train-dialogs-tags">
                     {trainDialog.tags.length === 0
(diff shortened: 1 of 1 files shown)


mattmazzola:
Fixes ADO#1924

LarsLiden:
Agreed.  The all read tags are too harsh.  I’d say just leave the tags in their default color (no red)

mattmazzola:
Ok, based on Lars feedback I changed it so tags don't change color, but I still change the empty text color.

image

PR #1026, opened 2019-03-05
Test Case Refactor + Scenario and Tags

SDESkowronski:
I refactored one of my existing test cases into a suite and am also including a new test suite "Scenario and Tags". The later is not finished; it has 6 stages complete with 4 stages still needing code.

The goal of this was to come to a common point of understanding between Matt and myself and any other team member who wishes to participate in the planning of how tests will be written going forward.

image

image


cypress/integration/Train/ScenarioAndTags.spec.js
```diff
@@ -0,0 +1,149 @@
+const models = require('../../support/Models')
```
pnwbrian: fyi, i'm pretty certain another reviewer has switch incoming test code to use imports and will make a similar note in his code review

SDESkowronski: Thanks, I fixed it.

cypress/integration/Train/ScenarioAndTags.spec.js
```diff
@@ -0,0 +1,149 @@
+import * as models from '../../support/Models'
+import * as modelPage from '../../support/components/ModelPage'
+import * as editDialogModal from '../../support/components/EditDialogModal'
+import * as train from '../../support/Train'
+import * as helpers from '../../support/Helpers'
+
+describe('Training Scenario and Tags', () => {
```
mattmazzola: Similar to the question in teams. Why do we need this test?  It looks like it's doing very similar things to the existing test and doesn't add missing cases or increase coverage

cypress/integration/Train/ScenarioAndTags.spec.js
```diff
@@ -0,0 +1,149 @@
...
+  context('Train Dialogs', () => {
+    context('Create', () => {
+      it('Should have no Scenario nor Tags when creating new dialog.', () => {
+        // Create new train dialog
+        // Verify that tags and description are empty
+        modelPage.NavigateToTrainDialogs()
+        //cy.WaitForTrainingStatusCompleted()
+        train.CreateNewTrainDialog()
+        
+        editDialogModal.VerifyScenario('')
```
mattmazzola: I don't want to have wrappers if we don't need them. We had talked about minimizing use of this earlier.

- Internally these use the custom `Get` instead of `get` which we should avoid.  Again want to use as low level tools and directly use Cypress. If we must wrap, why?  If their get was so unreliable they wouldn't have a designed it that way. Need to find out why it's not working.

- We want Intellisense which the helpers don't provide. The selectors were already managed in a single location so moving them to a helper doesn't improve that.

- A lot of the existing helpers were performing unneeded assertions or were coupled with other helpers that were required to be called first in order for some global to be the correct state.  We wanted to go a new direction to get away from all that.

cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js
```diff
@@ -26,11 +26,11 @@ describe('ErrorHandling', () => {
     editDialogModal.ClickSaveCloseButton()
 
     modelPage.VerifyErrorIconForTrainDialogs()
-    train.VerifyErrorsFoundInTraining(`${String.fromCharCode(59412)}Hey`, 'world peace', "Sorry $name, I can't help you get $want")
+    train.VerifyErrorsFoundInTraining(`${common.errorIconCharacter}Hey`, 'world peace', "Sorry $name, I can't help you get $want")
 
     // - - - Open the same Train Dialog, validate and fix the errors. - - -
 
-    train.EditTraining(`${String.fromCharCode(59412)}Hey`, 'world peace', "Sorry $name, I can't help you get $want")
+    train.EditTraining(`${common.errorIconCharacter}Hey`, 'world peace', "Sorry $name, I can't help you get $want")
```
mattmazzola: Glad to see this gone.  Brings up related question from previous review.

It would be more appropriate to separate the assertion of first input from the validity. There is a dedicated selector `data-testid="train-dialogs-validity-indicator"` to that was added specifically for this.  It should find a dialog and assert it's validity. Not find a first input and assume it's invalid if the first input starts with some character.

From my understanding the icon can be present even when its not an error just other validity cases so I would prefer to keep naming of validityIcon.

cypress/integration/Train/ScenarioAndTags.spec.js
mattmazzola: The title makes it sound like your going to learning/training the scenario.  It's just manually typed and edited by a human, no training involved. I would remove the word or make it Train Dialog since that's what we call it.

SDESkowronski: done

cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js
SDESkowronski: okay

cypress/integration/Train/ScenarioAndTags.spec.js
SDESkowronski: What question in teams? Please use teams to point it out to me. Thanks.

SDESkowronski: What existing test? I am not aware of an existing test for this scenario. This is the test you gave me in skeleton form with the beginning stages completed.

cypress/integration/Train/ScenarioAndTags.spec.js
SDESkowronski: `cy.Get`, `cy.Contains` and `cy.Click` are all wrappers around similar Cypress commands that start with a lowercase letter. The purpose of these are to avoid having to code explicit `cy.wait` all over the code. There are MANY reasons this is superior to using `cy.wait`.

1. `cy.wait` is unnecessary 99% of the time because we have code that monitors the DOM and knows definitively when it has completed rendering. If we ever come into a situation where we cannot tell this then 1st the product side should attempt to provide an indicator for such special cases so that we can reliably determine this. Because this is the greatest reason tests fail in other projects, but not ours! I am very proud of this fact.

2. Having `cy.wait` in the code is distracting when reading code...especially since they are unnecessary.

3. Writing code with `cy.wait` is very difficult to do, it adds 2-5 times more effort into writing a test case, and continues to add unnecessary extra effort when tests fail and people start tweaking the wait time to get the tests to pass.

4. You cannot get `cy.wait` right even 70% of the time. I've seen this issue over and over in nearly every test project I have been on...except the ones that were doing something similar to the DOM monitoring that I have done in our project.

cypress/integration/Train/ScenarioAndTags.spec.js
mattmazzola: https://github.com/Microsoft/ConversationLearner-UI/blob/master/cypress/integration/scenarioAndTags.spec.js

Also, it's not specific to Train Dialogs. It has some assertions about not showing tags or description on Log Dialogs too which is why it doesn't fit in that category.

cypress/integration/Train/ScenarioAndTags.spec.js
mattmazzola: Somehow this conversation turned into not using `cy.wait` when it started as not using `cy.Get` or other custom methods that wrap cypress.

> code that monitors the DOM 

That code seems to monitor the whole DOM not just the part we care about.  The DOM in other places can change without affecting the ability to get another element from stable part of the DOM.  E.g. It should be ok to click save on a dialog even though the training status is still changing.  This just how a user would act as well.  In this way the test could continue without waiting until the DOM has settled which would be faster.

> Having `cy.wait` in the code is distracting

Agreed, should be avoided. In most cases the request waiting should be enough.  E.g. No point in looking for a new train dialog until the request to add the new dialog or reload dialogs has returned.

> Writing code with cy.wait is very difficult to do, it

Hmm, I think the point is not about difficulty it's about efficiency.  Writing `cy.wait(N)` is easy. It just results in inefficient tests. We would rather tests only wait the required amount of time instead of some hard coded amount that is too long for most cases. Anyways, I'm not recommending it, just don't want all the wrappers. Especially ones that go into jQuery.

> You cannot get cy.wait right even 70% of the time.

What do you mean by "right"? 
You shouldn't have to get it right, this is why the commands implicitly wait until first success and do the waiting for you so you don't have to get it right.

SDESkowronski: Wrappers = Code Reuse = A Good Thing --- In my mind that is.

> I don't want to have wrappers if we don't need them. We had talked about minimizing use of this earlier.

Actually you talked about them and I ignored the subject at the time because I wanted to focus on other things and I was already over my head in changes you wanted me to make that I was having trouble digesting. I did not agree with you then and I don't agree with this statement you made.

In the case of `editDialogModal.VerifyScenario('')`...this is to avoid writing duplicate code about 6 times in this particular code file. And that is pretty much the same story for all of my wrappers/helper code. I've never been on any project, product nor test, that told me that wrapper functions and code reuse was a bad thing so I really don't understand where you are coming from on this.

They also contribute to code readability at the test case/spec level. The wrappers are named in such a way that they describe the step being taken in the test case. It makes the test cases self-documenting, to even non-developers...if the wrapper names need improvement for greater clarity I have no objection.

> A lot of the existing helpers were performing unneeded assertions

How is an assertion in test code unneeded? Even if an assertion is covered in another test case, having them in additional test cases or sub-scenarios increases the chance of catching edge case bugs...and they have done just that in this project and other projects I have been on that also are coded in this way (which would be all of them).

SDESkowronski: Please carry this forward from email.

cypress/integration/Train/ScenarioAndTags.spec.js
SDESkowronski: So that file was renamed/deleted so that it fit the naming pattern of all other test spec files.

Actually Log Dialog test stages was something I wanted to ask you, but we've been too busy debating other aspects of test code that I just did not get around to it...

Why are we looking in Log Dialogs for something that should not be there? Is there the potential in the UI product code that could put those fields in the Log Dialog view if there were some bug introduce there?

mattmazzola: Yes, there is potential for a bug to be introduced. The modal/window used for viewing dialogs is shared between train dialogs and log dialogs so we need to be sure those fields don't show up in certain circumstances such as when viewing a log dialog.

> So that file was renamed/deleted

Yes... and you deleted it.  I have another PR that was going to adjust it so I guess I'll put it back. At the time when I created the link the file existed on `master`, but you merged to master affecting the link. I should have linked to a fixed version:

https://github.com/Microsoft/ConversationLearner-UI/blob/1973bcf0a30a3d39a15cc45ffff534317c9edd38/cypress/integration/scenarioAndTags.spec.js

PR #1030, opened 2019-03-06
Refactor Create Tag and Frog

SDESkowronski:
While investigating a bug, I found it easier to pinpoint the exact point of failure by converting this code to our new spec style. Turns out it was bug [Bug 1901: Training Status is Misleading Causes Automatic Entity Labeling to NOT be Consistent](https://dialearn.visualstudio.com/BLIS/_workitems/edit/1901).

I also started making use of the "Scenario" and "Tags" fields of the Train Dialog since the grid looks weird without using those fields and we need to start converting all our tests to begin using them.


cypress/integration/CreateModels/TagAndFrog.spec.js
```diff
@@ -11,59 +11,87 @@ import * as editDialogModal from '../../support/components/EditDialogModal'
...
+      modelPage.NavigateToTrainDialogs()
+      cy.WaitForTrainingStatusCompleted()
+      train.CreateNewTrainDialog()
+    })
 
     // ------------------------------------------------------------------------
     // This block of code should be removed once we determine and fix the cause
     // of: Bug 1901-Automatic Entity Labeling Is NOT Consistent
     // ------------------------------------------------------------------------
+    it('Create a SPECIAL Training Dialog to deal with bug 1901', () => {
```
mattmazzola: These titles should explain what is being asserted not what is being done.  It's less important how it's done than if it is done correctly.

E.g. "should save the tags and description on the new dialog"

Then in the test you can clearly see how that is true:
```
                cy.get(testSelectors.trainDialogs.scenarios)
                    .should('contain', testData.description)

                cy.get(testSelectors.trainDialogs.tags)
                    .should('contain', testData.tag01)
```

cypress/integration/Train/ScenarioAndTags.spec.js
```diff
@@ -79,7 +79,7 @@ describe('Scenario and Tags - Train Dialog', () => {
         // reload
         // Verify edited tags and description are still on dialog
         editDialogModal.TypeScenario('Edited Test Scenario')
-        editDialogModal.AddTag('TagY')
+        editDialogModal.AddTags('TagY')
```
mattmazzola: Hmm, less important that the other things about test structure but if you're renaming this to be plural I would expect the method to take an array or multiple arguments.
Meaning you should haven't change the type of arguments depending on whether you want to add 1 or more.

This means the method should take:
```
addTags('tag01', 'tag02', 'tag03', ...)
addTags(['tag01', 'tag02', 'tag03'])
```
The latter form is more natural but I would be ok with former

cypress/support/components/EditDialogModal.js
```diff
@@ -42,9 +42,10 @@ export function VerifyTags(tags) {
   })
 }
 
-export function AddTag(tag) { 
-  cy.Get('button.cl-tags__button-add#tags').Click()
-  cy.Get('input#tags').type(`${tag}{enter}`)
+export function AddTags(tags) { 
+  ClickAddTagButton()
+  if (!Array.isArray(tags)) tags = [tags]
+  tags.forEach(tag => {cy.Get('[data-testid="tags-input-tag-input"]').type(`${tag}{enter}`)})
```
mattmazzola: `cy.Get('[data-testid="tags-input-tag-input"]')` Was this needed?
I know it's better to use `data-testid`s than classes but I thought it should have worked without re-finding the DOM element for each tag. The component is supposed to preserve focus on the input after pressing `{enter}` to allow the user to enter multiple easily.

mattmazzola:
> I found it easier to pinpoint the exact point of failure by converting this code to our new spec style.

Glad you see the benefits.

cypress/integration/Train/ScenarioAndTags.spec.js
SDESkowronski: This function does take either a single parameter or an array. Internally it converts to an array if a single string was passed in, and then proceeds to process it. So either of these calling syntaxes work:

> addTags('tag01')
addTags(['tag01', 'tag02', 'tag03'])
>

mattmazzola: Yes, I know I see the line: `if (!Array.isArray(tags)) tags = [tags]`
https://github.com/Microsoft/ConversationLearner-UI/pull/1030/files#diff-bff231d2b160a6454f2005358f643af4R47

> you should haven't change the type of arguments

Yes, going from string to array would be considered changing the type.

cypress/support/components/EditDialogModal.js
SDESkowronski: Good catch, I was able to improve this. I've updated the code with the change.

cypress/integration/Train/ScenarioAndTags.spec.js
SDESkowronski: Working on it...

SDESkowronski: Updates to `cease changing the type` in this case and elsewhere have been made.

cypress/support/components/EditDialogModal.js
```diff
@@ -42,10 +42,15 @@ export function VerifyTags(tags) {
...
-  cy.Get('button.cl-tags__button-add#tags').Click()
-  cy.Get('input#tags').type(`${tag}{enter}`)
-  cy.WaitForStableDOM()
+// Pass in either an array of tag strings or a single tag string.
+// If you try to call this twice in a row, it will fail to find the "Add Tag Button"
+// so don't do it, this was designed to take multiple tags.
+export function AddTags(tags) { 
+  ClickAddTagButton()
+  let tagList = ''
+  if (!Array.isArray(tags)) { tagList = `${tags}{enter}` }
```
mattmazzola: Doesn't look like it would solve the issue about switching types. E.g. string to array, it just changed the implementation depending on if it is a string or array. This now worse than before because it generates the`tagList` in two places conditionally.  I don't know how to make myself more clear:

> you should haven't change the type of arguments depending on whether you want to add 1 or more.

If at one point I write a test that adds 1 tag: `AddTags('MyTag')` and then I realize later it should add 2 tags I should NOT have to change the type. Meaning just add a tag: `AddTags('MyTag', 'MySecondTag')` The argument types stays as strings.

You either take multiple arguments as strings and array of strings.  Could do both but that's more complicated and would rather people use the method the same way every time.  The having mixed and people being confused.

SDESkowronski: The other change I made was because I was thinking your main objection was changing the type of a function argument. Sorry I misunderstood. I fixed this. Need to re-review the code. Impacted many places.

I actually liked it the other way, it was like function/method overloading in other languages. I thought other people would like it too. Clearly you don't so it is gone. They now all take arrays, which I think is awkward/unnecessary when you pass in a single item as is the most common use case with these functions.

cypress/support/components/EditDialogModal.js
SDESkowronski: STILL TODO

PR #1031, opened 2019-03-07
fix: preserve source dialog instead of new dialog when continuing

mattmazzola:
When continuing instead of editing the original it would remove the original and add the continued/new dialog which was created as side effect of the new teach session.

This caused the created time to update when it shouldn't have.

The other problem is much more complicated but related:
Due to the way the thunk was written, in combination with other thunks and the asynchronous `await`s inside it, the call to delete the source dialog would finish and remove the dialog from the list before the next call refresh the dialogs which implicitly added it to the list.

Due to created date being changed this caused the dialog to jump/shift in place in the list and depending on server response times could look like the dialog was deleted before it would reappear.

After all is done the dialog properly updates in place and preserves the create date.

Fixes ADO#1907
Might fix: ADO#1905 (Still need to confirm this)

### DIFF
--- a/src/actions/teachActions.ts
+++ b/src/actions/teachActions.ts
@@ -9,7 +9,7 @@ import * as ClientFactory from '../services/clientFactory'
 import { setErrorDisplay } from './displayActions'
 import * as CLM from '@conversationlearner/models'
 import { AxiosError } from 'axios'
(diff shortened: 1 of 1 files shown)


mattmazzola:
I made some videos so you can see the problem:

Existing behavior with Removal:
http://recordit.co/GSiVFL1OnM

With Fix:
http://recordit.co/oRKzu52Ibp

PR #1033, opened 2019-03-08
fix: improve entity label conflict modal

mattmazzola:
Adds additional information to entity label conflict modal.  Aims to improve the following

- Shows the current labels they attempted to submit which were rejected due to conflicts so they can compare it with previously submitted labels and see the problem
- Explains why the bot needs consistent labelling
- Explains what clicking Accept will do.

image

Future improvements:
Out of scope for PR and I had already talked to Lars about it but I think we eventually should let the user continue even if the labels cause a conflict.  If they had labeled the input incorrectly the first time and now try to label the input correctly this time they can't and are stuck. They won't want to accept changes because they're wrong and closing the dialog does nothing.  They could either stop the dialog or accept the incorrect labels and then remember to go back and update all the incorrect labels.  Being able to continue from here is likely not easy operation either.  The service would have to implicitly update all the existing dialogs which now have conflicts to use the new labels and and still ensure they can be replayed correctly with different set of entities labeled.

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -69,6 +69,10 @@ html {
   color: var(--color-alert);
 }
 
+.cl-text--success {
+  color: var(--color-success);
+}
+
 .cl-action {
   border: none;
   background: none;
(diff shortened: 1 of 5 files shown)


mattmazzola:
Ok, left the "current" part since Lars thought it was important for the user to know that it could be from the same dialog but I updated the subtitle to match your suggestion

PR #1034, opened 2019-03-12
fix: bug 1941 - add testid attributes to label conflict modal

pnwbrian:
Adds data-testid attributes to the new entity label conflict modal to support testing

No change to product level functionality

### DIFF
--- a/src/components/modals/ExtractConflictModal.tsx
+++ b/src/components/modals/ExtractConflictModal.tsx
@@ -33,6 +33,7 @@ const ExtractConflictModal: React.SFC<Props> = (props) => {
             className={OF.FontClassNames.mediumPlus}
             containerClassName="cl-modal cl-modal--small"
             onDismiss={() => props.onClose()}
+            data-testid="entity-labels-conflict-modal"
         >
             <div className={`cl-modal_header cl-text--error ${OF.FontClassNames.xLarge} `}>
                 <OF.Icon iconName="Warning" />&nbsp;{formatMessageId(intl, FM.EXTRACTCONFLICTMODAL_TITLE)}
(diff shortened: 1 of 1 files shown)


mattmazzola:
What bug does this fix? 

I would prefer to see these with the test that uses them to know they are appropriate and needed otherwise we just add data-testid to everything and it becomes hard to know which is used.

src/components/modals/ExtractConflictModal.tsx
```diff
@@ -42,7 +43,7 @@ const ExtractConflictModal: React.SFC<Props> = (props) => {
                 <p>{formatMessageId(intl, FM.EXTRACTCONFLICTMODAL_SUBTITLE)}</p>
                 <div>{formatMessageId(intl, FM.EXTRACTCONFLICTMODAL_REVIEW)}</div>
 
-                <div className="cl-inconsistent-entity-modal-header cl-text--error"><OF.Icon iconName="ChromeClose" />&nbsp;{formatMessageId(intl, FM.EXTRACTCONFLICTMODAL_CONFLICTING_LABELS)}</div>
+                <div className="cl-inconsistent-entity-modal-header cl-text--error" data-testid="entity-labels-conflict-submitted"><OF.Icon iconName="ChromeClose" />&nbsp;{formatMessageId(intl, FM.EXTRACTCONFLICTMODAL_CONFLICTING_LABELS)}</div>
```
mattmazzola: These wording seems misleading / backwards to me.  Prefer to keep the ids to logically match the graphic;
image

E.g.  Conflicting vs Previously Submitted

The top input which was rejected due to the conflicts, the bottom was previously submitted and accepted.  Technically they are both submitted but one only succeeded so submitted word doesn't differentiate

pnwbrian: I'm creating new test automation code against this new functionality, and need a direct way to access these UI elements in keeping with our existing test code (vs walking the DOM).

mattmazzola: Right... we use `data-testid` from the tests. I know why you're doing it but disagree with how/when you're doing it.

PR #1036, opened 2019-03-13
fix: show correct number of log dialogs

mattmazzola:
On the actual log dialogs page we were excluding dialogs that had been saved as train dialogs.

The meant the count displayed would be incorrect.
There was also bug of the logic to show the empty list placeholder list using a different set of items than the actual list.

Fixes ADO#1943

### DIFF
--- a/src/routes/Apps/App/Index.tsx
+++ b/src/routes/Apps/App/Index.tsx
@@ -192,6 +192,7 @@ class Index extends React.Component<Props, ComponentState> {
 
         const trainDialogValidity = this.getTrainDialogValidity();
         const invalidBot = this.state.botValidationErrors && this.state.botValidationErrors.length > 0;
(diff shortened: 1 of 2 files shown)


LarsLiden:
LogDialogs saved as TrainDialogs should be deleted when the TrainDialog is created.

mattmazzola:
Yes, they should be but either way this code is still needed.

I actually was able to save the dialog. I thought this occurred when I attempted to save it.

mattmazzola:
Adding stuff Lars and I talked about in person

> Should we delete all converted log dialog?
No, we can't because it might still be needed.  If the conversion is within 30 minutes or before End Session the user can still continue it even though the train dialog created from it is only partial.

> If we preserve how should indicate this to the user?
Do preserve on server but still filter them out of view. Means we don't need new indicator.

PR #1037, opened 2019-03-13
chore: use const when applicable

mattmazzola:
Similar to #1035 but covers the /src folder instead of /cypress.  Doesn't change functionality but cleans up code to use `const` when possible to prevent accidental reassignment and make it easier to read.

Would be more work to go through the remaining `let`s to see if they can be removed but the benefit isn't worth the time.


LarsLiden:
Were you able to do this in an automate fashion?

mattmazzola:
No, but it's close.  Since it doesn't change the logic it was basically a replace call then let the fix the compiler errors.

PR #1039, opened 2019-03-14
Fix Train Tag and Frog

SDESkowronski:
Test case was broken due to recent product changes.


cypress/integration/Train/TagAndFrog.spec.js
```diff
@@ -8,72 +8,92 @@ import * as modelPage from '../../support/components/ModelPage'
 import * as train from '../../support/Train'
 import * as editDialogModal from '../../support/components/EditDialogModal'
 import * as common from '../../support/Common'
+import * as helpers from '../../support/Helpers'
+
+describe('Tag And Frog - Train', () => {
+  // TODO: Need to add another test case or expand this one so that tagging something
+  //       that was NOT tagged in another instance causes the UI to complain.
```
SDESkowronski: Hey Brian, this TODO comment is a good candidate for enhancing this test case next time you are looking for work to do.

cypress/integration/Train/TagAndFrog.spec.js
```diff
@@ -8,72 +8,92 @@ import * as modelPage from '../../support/components/ModelPage'
...
+  const textEntityPairs = [{ text: 'Tag', entity: 'multi' }, { text: 'Frog', entity: 'multi' }]
+
+  afterEach(helpers.SkipRemainingTestsOfSuiteIfFailed)
+  
+  context('Setup', () => {
```
mattmazzola: Since this is setup it looks like a would be more appropriate to use `before`

cypress/integration/Train/TagAndFrog.spec.js
```diff
@@ -8,72 +8,92 @@ import * as modelPage from '../../support/components/ModelPage'
...
+      editDialogModal.ClickScoreActionsButton()
+      editDialogModal.VerifyEntityLabeledDifferentPopupAndClose(textEntityPairs)
+      editDialogModal.ClickScoreActionsButton()
+      editDialogModal.VerifyEntityLabeledDifferentPopupAndAccept(textEntityPairs)
+      train.SelectAction('Hi')
+    })
+  })
 
-    train.Save()
+  context('Setup for next phase', () => {
```
mattmazzola: Similar to above. Given title says "Setup" I would expect this isn't an actual test and could be included the previous test `after` or next `before` sections

cypress/integration/Train/TagAndFrog.spec.js
SDESkowronski: Well, technically it is a test. It is running the UI in order to do setup work, it can and does fail from time to time, and it is validating the UI along the way.

Originally I did put the setup code in a `before` function, but I did not like how it showed up in the UI, in some ways it was hidden, and in other ways it was too much in the way and awkward to minimize to get the overall picture of the suite progress. 

**These first set of images show how it looks, and the steps I have to take to get to a summary view of the test suite run...**

image

**So I have to expand the first test in order to see the setup phase, but I then get this...**

image

**Followed by a very long list of events that occurred. So then you have to click the icon to minimize that, which then gives you this view...**

image

**Again followed by a very long list of events that occurred. So then you have to click the icon to minimize that, which then gives you this view...**

image

**Which brings me to the best summary view of it I can get (see above image), which I don't like because the 'BEFORE ALL' part really should appear above 'Train - Standard Input' part and it does not explain what is happening in that 'BEFORE ALL' like it can be explained when that code is run as part of an 'it()' function.**

**HOWEVER, when I have the setup steps in a test, I get a nice summary view like the following without having to do any extra clicking around and the setup phases are nicely explained...**

image

PR #1040, opened 2019-03-15
chore: add icons to buttons

mattmazzola:
I noticed some places our buttons have icons and other places they don't so I wanted to see what it looked like it we added them in more places.

Having a visual indicator can be more well-known than language so this might help non-English speakers.

Some of them of simple like, adding + indicator when we create something, but there are many others.
I tried to create image of before / after to see. 

There is still an issue we have to figure out with icon contrast when the button is disabled, so this will have to wait until that is fixed, but wanted to see what people felt about change visually.

image

There are some other buttons that change the action depending on context such as TeachSessionModal depending if you delete or abandon or continue etc.  I tried to leave those alone for now, but let me know if I have missed some or icon isn't semantic to the button.


mattmazzola:
Ok, seems to have fixed issue with disabled buttons:
image

PR #1043, opened 2019-03-16
Many Bug Fixes and Test Spec Refactor

SDESkowronski:
- I fixed multiple bugs in test cases - All known test bugs are fixed now!
- Along the way I refactored the test spec files that I was debugging into our new pattern.
- I found some more globals hanging off of the "window" object, I changed to using "let"


cypress/integration/EditAndBranching/TagAndFrog.spec.js
```diff
@@ -8,34 +8,46 @@ import * as modelPage from '../../support/components/ModelPage'
...
+    it('Should verify user cannot submit changes without accepting auto-re-labling of those two entity labels in the alternative input', () => {
+      editDialogModal.ClickSubmitChangesButton()
+      editDialogModal.VerifyEntityLabeledDifferentPopupAndClose(textEntityPairs)
+      editDialogModal.ClickSubmitChangesButton()
+      editDialogModal.VerifyEntityLabeledDifferentPopupAndAccept(textEntityPairs)
 
-    editDialogModal.VerifyEntityLabeledDifferentPopupAndClose(textEntityPairs)
-    editDialogModal.ClickSubmitChangesButton()
-    editDialogModal.VerifyEntityLabeledDifferentPopupAndAccept(textEntityPairs)
+      editDialogModal.VerifyEntityLabeledDifferentPopupAndClose(textEntityPairs)
```
mattmazzola: Why does this pop up again, if they accepted the changes in the above and it's the same `textEntityPairs` ?  I would expect the conflict to be resolved and the window should not pop up any more.

cypress/integration/EditAndBranching/TagAndFrog.spec.js
```diff
@@ -8,34 +8,46 @@ import * as modelPage from '../../support/components/ModelPage'
...
-    editDialogModal.SelectChatTurnExactMatch('This is Tag.')
+  context('Edit Train Dialog', () => {
+    it('Should remove two entity labels from alternative input', () => {
+      train.EditTraining('This is Tag.', 'This is Tag.', 'Hi')
+      editDialogModal.SelectChatTurnExactMatch('This is Tag.')
 
-    editDialogModal.VerifyEntityLabelWithinSpecificInput([textEntityPairs[0]], 0)
-    editDialogModal.VerifyEntityLabelWithinSpecificInput(textEntityPairs, 1)
-    editDialogModal.VerifyEntityLabelWithinSpecificInput(textEntityPairs, 2)
+      editDialogModal.VerifyEntityLabelWithinSpecificInput([textEntityPairs[0]], 0)
```
mattmazzola: Why does the first call use `[textEntityPairs[0]]` and not the rest?  I see some weird logic around indices of slate editor but looks unecessary.

cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js
```diff
@@ -26,11 +26,11 @@ describe('ErrorHandling', () => {
     editDialogModal.ClickSaveCloseButton()
 
     modelPage.VerifyErrorIconForTrainDialogs()
-    train.VerifyErrorsFoundInTraining(`${common.validityIcon}Hey`, 'world peace', "Sorry $name, I can't help you get $want")
+    train.VerifyErrorsFoundInTraining(`Hey`, 'world peace', "Sorry $name, I can't help you get $want")
```
mattmazzola: Unrelated to review, but eventually find a way to share these as constant test data object.

cypress/integration/Train/MyNameIs.spec.js
```diff
@@ -30,7 +30,10 @@ describe('Train', () => {
     cy.WaitForTrainingStatusCompleted()
 
     train.TypeYourMessage('My name is Susan.')
+// TODO: Remove these left indented lines once the product bug is fixed.    
+cy.log('Going to hit bug 1948')
```
mattmazzola: Is this not fixed by adding the extra input into the dialog?  Shahin said he tested it yesterday, I thought it fixed it.

cypress/support/MonitorDocumentChanges.js
```diff
@@ -34,6 +35,16 @@ import * as helpers from './Helpers.js'
 
   function MillisecondsSinceLastChange() {
     if (new Date().getTime() > lastMonitorTime + 50) LookForChange()
+    
+    // Some operations are known to take a long time without rendering, so while this count
+    // is greater than zero, we will reset our lastChangeTime and return zero which will
+    // effectively cause a wait until the specified number of changes pass and then resume
+    // our normal monitoring for stable DOM.
+    if (waitTillNChangesOccur > 0) { 
+      lastChangeTime = new Date().getTime()
```
mattmazzola: What are the cases this is needed?  Would rather use the builtin Cypress `timeout` instead as this is what it's for.

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -125,9 +125,9 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
                         <span data-testid="train-dialogs-description">
                             {trainDialog.description
                                 || <>
-                                    <span data-testid="train-dialogs-first-input">{firstInput ? firstInput : ''}</span>
+                                    <span /*data-testid="train-dialogs-first-input"*/>{firstInput ? firstInput : ''}</span>
```
mattmazzola: Good to see this. Minor, Would prefer to remove the comment also.

cypress/integration/ErrorHandling/MissingAction.spec.js
```diff
@@ -10,44 +10,66 @@ import * as editDialogModal from '../../support/components/EditDialogModal'
...
+  context('Setup', () => {
+    it('Should import a model and wait for training to complete', () => {
+      models.ImportModel('z-missingAction', 'z-whatsYourName.cl')
+      modelPage.NavigateToTrainDialogs()
+      cy.WaitForTrainingStatusCompleted()
+    })
 
+    it('Should verify there are no error icons on the page and create a new Train Dialog', () => {
+      modelPage.VerifyNoErrorIconOnPage()
     train.CreateNewTrainDialog()
```
mattmazzola: Minor, but it's slightly more appropriate to be grouped with the next test which modifies the dialog

cypress/integration/ErrorHandling/MissingAction.spec.js
```diff
@@ -10,44 +10,66 @@ import * as editDialogModal from '../../support/components/EditDialogModal'
...
-
-    modelPage.VerifyErrorIconForTrainDialogs()
-    train.VerifyErrorsFoundInTraining(`${common.validityIcon}${common.gonnaDeleteAnAction}`, common.gonnaDeleteAnAction, undefined)
-
-    train.EditTraining(`${common.validityIcon}${common.gonnaDeleteAnAction}`, common.gonnaDeleteAnAction, undefined)
+  context('Train', () => {
+    it('Should complete and save a simple 1 action Train Dialog', () => {
+      train.TypeYourMessage(common.gonnaDeleteAnAction)
+      editDialogModal.ClickScoreActionsButton()
+      train.SelectAction(common.whatsYourName)
```
mattmazzola: This actually all setup for the actual test which actually verifies what that deleting an action causes the dialog to become invalid (contain errors due to missing action).

That should be reflected in the test hierarchy.  These are just extra `it`s not really testing anything.  Would rather this all be setup.

cypress/integration/ErrorHandling/MissingAction.spec.js
```diff
@@ -10,44 +10,66 @@ import * as editDialogModal from '../../support/components/EditDialogModal'
...
+    it('Should delete the action we just used in the Train Dialog', () => {
+      modelPage.NavigateToActions()
+      actions.DeleteAction(common.whatsYourName)
+    })
+  })
 
-    editDialogModal.SelectChatTurnStartsWith('ERROR: Can’t find Action Id')
-    editDialogModal.VerifyErrorMessage('Action does not exist')
-    scorerModal.VerifyMissingActionNotice()
+  context('Train Dialog Grid', () => {
```
mattmazzola: Would prefer this to be in plural form "Train Dialogs" can add "Grid" if you want to make it clear it's the list view.  There are other lists/grids for the memory and action selection on the train dialog so this is confusing to me.

One is verifying the error icon in the list view and the other is verify the more specific error content in the modal.

cypress/integration/ErrorHandling/MissingAction.spec.js
```diff
@@ -10,44 +10,66 @@ import * as editDialogModal from '../../support/components/EditDialogModal'
...
-    editDialogModal.VerifyNoErrorMessage()
+      editDialogModal.SelectChatTurnStartsWith('ERROR: Can’t find Action Id')
+      editDialogModal.VerifyErrorMessage('Action does not exist')
+      scorerModal.VerifyMissingActionNotice()
+    })
+  })
 
-    editDialogModal.ClickSaveCloseButton()
-    modelPage.VerifyNoErrorIconOnPage()
+  context('Train Add Action', () => {
```
mattmazzola: Looks like it's continue from next text so it's also Train Dialog text, Would put it with the one above.

cypress/support/Train.js
```diff
@@ -8,15 +8,18 @@ import * as trainDialogsGrid from './components/TrainDialogsGrid'
 import * as editDialogModal from './components/EditDialogModal'
 import * as helpers from './Helpers'
 
-function Today() { return Cypress.moment().format("MM/DD/YYYY") }
+let currentTrainingSummary
+let originalTrainingSummary
+let isBranched
+let originalChatMessages
+let editedChatMessages
```
mattmazzola: Better than using `window` but I think this still has the same affect of creating a global.

cypress/integration/EditAndBranching/TagAndFrog.spec.js
SDESkowronski: In this test case we create two errors in two different alt-inputs. So after we fix one of them we have the other one popup. We first close it to confirm that we can not get away with not fixing it, then we follow up and accept it.

cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js
SDESkowronski: That is what I thought I did, but when we meet up on Wednesday you can explain to me how to do it the way you like it. I did see how you defined the constants in the "Description and Tag" test case, perhaps that is what you are referring to...and I am not currently following that model, but I can, I have no objection to that.

cypress/integration/Train/MyNameIs.spec.js
SDESkowronski: I will try that, haven't tried yet...if I can get it working I'll remove the log and comment.

cypress/support/MonitorDocumentChanges.js
SDESkowronski: [Bug 1950](https://dialearn.visualstudio.com/BLIS/_workitems/edit/1950): Save Train Dialog takes too long to render grid results without spinner nor discernable API call activity

So this is fudging for what I believe is a product bug. Which is why I filed the bug. 

Saving the Train Dialog has consistently had odd rendering issues from the time I started testing it. They get fixed, only for a new one to crop up a few weeks later with new changes.

The reason I like this solution is that it is definitive and precise. We can see how many changes constitute a render in the logs, and just wait for them to occur before we resume normal operations. Also it is implemented at a single step of a test, saving the Train Dialog, as opposed to the other multiple possible points of failure.

The biggest reason is that it stays consistent of the philosophy and patterns behind the DOM Monitoring code, and that is to block test progress until we know the DOM is stable after whatever recent changes were triggered by UI activity. Last I heard you were objecting to this DOM Monitoring code so this too will need to be discussed when we have our meeting on Wednesday.

src/routes/Apps/App/TrainDialogs.tsx
SDESkowronski: Okay, I removed the comments.

cypress/integration/ErrorHandling/MissingAction.spec.js
SDESkowronski: Done.

cypress/integration/ErrorHandling/MissingAction.spec.js
SDESkowronski: Yep, that sounds about right. I fixed it.

cypress/integration/ErrorHandling/MissingAction.spec.js
SDESkowronski: It is a little different, but I see your point so I moved it up and modified the test description slightly to better describe what that stage is doing.

cypress/support/Train.js
SDESkowronski: I've done a lot of reading about let and globals and it seems that this is not creating a true global. It appears that the distinctions are subtle, but that it does not have the same pitfalls as the global can have.

However, we can discuss how you would like this sort of thing to be done when we meet up on Wednesday.

cypress/integration/EditAndBranching/TagAndFrog.spec.js
SDESkowronski: In this case there are three user inputs being manipulated:

1. The main user utterance with 'Tag' being labeled as 'multi' entity.
2. The 1st alternative input, with both entity pairs in that array being labeled.
3. the 2nd alternative input, with both entity pairs in that array being labeled.

As far as specifying the indices of the "slate editor"...Either I have to add logic to the Verify function that finds the user utterance based on a string, or I can specify it as an index. In this case it seemed better to use an index, but I think the spec code is easier to read and understand when it is a string. I would be happy to change it to a string if you also think that is better. Or do you have another suggestion on how to handle such a function?

cypress/integration/ErrorHandling/MissingAction.spec.js
SDESkowronski: Made it plural. 

I am not sure I understood the rest of your comment, but I'd be happy to reword anything else so as to keep the language consistent with whatever way the team refers to each component in question.

cypress/integration/Train/MyNameIs.spec.js
SDESkowronski: Shahin did fix it as long as I add the extra input. I'll fix that in the test code in my next go at it. Need to play around with it a bit more before I decide how I want to test this and if indeed something else was not at issue here that might have been missed.

PR #1046, opened 2019-03-19
Delete All Test Models For Parallel Test Runs

SDESkowronski:
Experimenting...CircleCI is not running my build so I am trying a PR to see if that makes it go.


.circleci/config.yml
```diff
@@ -51,7 +52,12 @@ jobs:
...
...ey $CYPRESS_RECORD_KEY --parallel --spec "cypress/integration/CreateModels/AllEntityTypes.spec.js,cypress/integration/CreateModels/DisqualifyingEntities.spec.js,cypress/integration/CreateModels/Travel.spec.js,cypress/integration/CreateModels/WaitVsNoWaitActions.spec.js,cypress/integration/CreateModels/WhatsYourName.spec.js,cypress/integration/EditAndBranching/AddEndSessionAction.spec.js,cypress/integration/EditAndBranching/Branching.spec.js,cypress/integration/EditAndBranching/VerifyEditTrainingControlsAndLabels.spec.js,cypress/integration/ErrorHandling/ActionUnavailable.spec.js,cypress/integration/ErrorHandling/MissingAction.spec.js,cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js,cypress/integration/ErrorHandling/WaitNonWait.spec.js,cypress/integration/Log/WhatsYourName.spec.js,cypress/integration/Train/BookMeAFlight.spec.js,cypress/integration/Train/DisqualifyingEntities.spec.js,cypress/integration/Train/MyNameIs.spec.js,cypress/integration/Train/ScenarioAndTags.spec.js,cypress/integration/Train/TagAndFrog.spec.js,cypress/integration/Train/WaitVsNoWaitActions.spec.js,cypress/integration/Train/WhatsYourName.spec.js,cypress/integration/UX/BotModelMismatch.spec.js"
```
mattmazzola: Why does it need to be written like this?

It's difficult to read and difficult to maintain

cypress/support/components/EditDialogModal.js
```diff
@@ -249,10 +249,12 @@ export function VerifyEntityLabeledDifferentPopupAndAccept(textEntityPairs) { Ve
 
 function VerifyEntityLabeledDifferentPopupAndClickButton(textEntityPairs, buttonLabel) {
   cy.Get('.ms-Dialog-main')     // This returns multiple parent objects
-    .contains('Entity is labelled differently in another user utterance') // Narrows it down to 1
-    .parents('.ms-Dialog-main') // Back to the single parent object
+    .contains('Inconsistent Entity Labels') // Narrows it down to the one we want
```
mattmazzola: Would seem better to change this selector to use the `[data-testid= ]` like this rest of selectors so it doesn't need to match the title.

.circleci/RevealAuthoringKeys.js
```diff
@@ -0,0 +1,5 @@
+console.log(`process.env.LUIS_AUTHORING_KEY_ALT_1=${process.env.LUIS_AUTHORING_KEY_ALT_1}\n`)
```
mattmazzola: The `SetLuisAuthoringKey.js` was already risky but accepted because output was piped to the $BASH_ENV

This `- run: node .circleci/RevealAuthoringKeys.js` is run directly which I think means it output is shown in the circle CI logs. We can't expose the environment variables to those logs I think these are also public.

.circleci/config.yml
SDESkowronski: These are changes in progress. I needed to test CircleCI and the only way I knew of doing that was by actually submitting a PR. It was not triggering build when I pushed changes into my branch. All of the garbage will be removed before final PR.

I thought if I did not assign you as a reviewer that you would ignore this.

.circleci/RevealAuthoringKeys.js
mattmazzola: You will have to rotate all these keys now

SDESkowronski: When you say rotate it sounds like I just need to re-order the keys in the array in SetLuisAuthoringKey.js file. I am okay with doing it that way, but I would assume I need to get whole new keys and delete the old ones...is that what you really meant?

mattmazzola: Yes, should have said regenerate all of them. rotate is for difference process when keys expire and you have to regenerate one of them.

> I would assume I need to get whole new keys and delete the old ones

Yes, they are leaked to public now.  Dont expose secrets

.circleci/config.yml
```diff
@@ -28,6 +28,7 @@ jobs:
 
       # Randomly pick 1 out of 5 LUIS Authoring Keys and set the BASH_ENV with it.
       - run: node .circleci/SetLuisAuthoringKey.js >> $BASH_ENV
+      - run: node .circleci/RevealAuthoringKeys.js
```
mattmazzola: Remove this line

.circleci/config.yml
```diff
@@ -50,8 +50,11 @@ jobs:
...
...-record --key $CYPRESS_RECORD_KEY --spec "cypress/integration/CreateModels/AllEntityTypes.spec.js,cypress/integration/CreateModels/DisqualifyingEntities.spec.js,cypress/integration/CreateModels/Travel.spec.js,cypress/integration/CreateModels/WaitVsNoWaitActions.spec.js,cypress/integration/CreateModels/WhatsYourName.spec.js,cypress/integration/EditAndBranching/AddEndSessionAction.spec.js,cypress/integration/EditAndBranching/Branching.spec.js,cypress/integration/EditAndBranching/VerifyEditTrainingControlsAndLabels.spec.js,cypress/integration/ErrorHandling/ActionUnavailable.spec.js,cypress/integration/ErrorHandling/MissingAction.spec.js,cypress/integration/ErrorHandling/TwoConsecutiveUserInput.spec.js,cypress/integration/ErrorHandling/WaitNonWait.spec.js,cypress/integration/Log/WhatsYourName.spec.js,cypress/integration/Train/BookMeAFlight.spec.js,cypress/integration/Train/DisqualifyingEntities.spec.js,cypress/integration/Train/MyNameIs.spec.js,cypress/integration/Train/ScenarioAndTags.spec.js,cypress/integration/Train/TagAndFrog.spec.js,cypress/integration/Train/WaitVsNoWaitActions.spec.js,cypress/integration/Train/WhatsYourName.spec.js,cypress/integration/UX/BotModelMismatch.spec.js"
```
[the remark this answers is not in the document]
SDESkowronski: Run in a folder, I'll keep that in mind and move to that soon.

cypress/support/components/EditDialogModal.js
SDESkowronski: Good point, I was able to significantly simplify this method. When I originally wrote it that data-testid was not there, and I could not add one or could not get it right (don't remember now). But now that it has changed, and you added that in I was able to make it better. Thanks.

SDESkowronski: The fix will be in another PR since this one has been pushed to master.

PR #1049, opened 2019-03-20
feat: utilities for finding matching train dialogs

LarsLiden:
Pre-work needed for being able to merge train dialogs
some additional code cleanup on treeview

### DIFF
--- a/src/Utils/dialogUtils.ts
+++ b/src/Utils/dialogUtils.ts
@@ -133,3 +133,80 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
     }
     return best
 }
+
+function doesScorerStepMatch(scorerStep1: CLM.TrainScorerStep, scorerStep2: CLM.TrainScorerStep): boolean {
+    if (scorerStep1.labelAction !== scorerStep2.labelAction) {
+        return false
+    }
+    if (scorerStep1.input.filledEntities.length !== scorerStep2.input.filledEntities.length) {
+        return false
+    }
+    let entityIndex = 0
+    while (entityIndex < scorerStep1.input.filledEntities.length) {
+        let entity1Id = scorerStep1.input.filledEntities[entityIndex].entityId
+        if (!scorerStep2.input.filledEntities.find(fe => fe.entityId === entity1Id)) {
+            return false
+        }
+        entityIndex = entityIndex + 1
+    }
+    return true
+}
+
+function doesRoundMatch(round1: CLM.TrainRound, round2: CLM.TrainRound, isLastRound: boolean): boolean {
+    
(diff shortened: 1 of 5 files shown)


src/Utils/dialogUtils.ts
```diff
@@ -133,3 +133,80 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
...
+        }
+        stepIndex = stepIndex + 1
+    }
+    return true
+}
+
+export function doesTrainDialogMatch(trainDialog1: CLM.TrainDialog, trainDialog2: CLM.TrainDialog): boolean {
+    const maxRounds = Math.max(trainDialog1.rounds.length, trainDialog2.rounds.length)
+    let roundIndex = 0
+    while (roundIndex < maxRounds) {
```
mattmazzola: Seems like this is equivalent to an `zip` and `every` call which, in my opinion, more clearly explains what it's doing.
`zip` is common function that pairs elements of array limited by smaller array together (pairing rounds)
`every` is common function that say every round must match

Something like:
```
return zip(trainDialog1.rounds, trainDialog2.rounds)
  .every(({ round1, round2 }, roundIndex, roundPairs) => {
     const isLastRound = (roundIndex === roundPairs.length - 1)
     return doesRoundMatch(round1, round2, isLastRound)
  })
```

`zip` isn't standard javascript yet though so many libraries provide it

src/Utils/testDataUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+        "userText": "userText",
+        "displayText": "displayText",
+        "builtinType": null,
+        "resolution": {}
+    }]
+}
+
+export function makeFilledEntity(entityId?: string): CLM.FilledEntity {
+    return {
+        entityId: entityId || CLM.ModelUtils.generateGUID(),
```
mattmazzola: Hmm, we already have the `uuid` library included on this repo which does a real guid.  Doesn't matter much though

src/Utils/testDataUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+        values: makeEntityValues()
+    }
+}
+
+export function makeScoreInput(entityIds?: string[]): CLM.ScoreInput {
+    const fillledEntities = entityIds ?
+        entityIds.map(eid => makeFilledEntity(eid)) : [makeFilledEntity()]
```
mattmazzola: Minor, but I prefer the results of conditional on separate lines for things more complicated than simple values.

Similar to what you did for `scoreSteps` below:
https://github.com/Microsoft/ConversationLearner-UI/pull/1049/files#diff-daf9ef0c30e15880734bf4a8ef171916R43

PR #1050, opened 2019-03-20
fix: explicitly set sourceLogDialogId when saving

mattmazzola:
Depends on server and models changes so will have to wait but at least we can show it now.

To recap, there is pending change in models so `sourceLogDialogId` is not implicitly set on all temporary train dialogs. This fixed the issue of implicitly removing the associated log dialogs as you edited them without saving, but it means we now need to explicitly set the `sourceLogDialogId` when clicking "Save As Train Dialog" so but sure the association is only created when the user actually saves changes.


mattmazzola:
Ended up doing a different change on server and models so this isn't needed anymore.
It now removes the temporary train dialog id from the list when session is deleted.  There wasn't any easy choice but overall it seemed safer this way. We're allowing log dialogs to get into bad state and then relying on cleaning the issue up instead of preventing the bad state, but the alternative was to make the source dialog ID mutable on train dialog update call which may have posed more risk related to GDPR.

PR #1051, opened 2019-03-20
feat: merge train dialogs and test

LarsLiden:
Just backend support code.  no UI yet

### DIFF
--- a/src/Utils/dialogUtils.ts
+++ b/src/Utils/dialogUtils.ts
@@ -4,6 +4,7 @@
  */
 import * as CLM from '@conversationlearner/models'
 import { Activity } from 'botframework-directlinejs'
+import { deepCopy } from './util'
 
 export interface DialogRenderData {
     dialogMode: CLM.DialogMode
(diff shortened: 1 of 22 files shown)


src/Utils/dialogUtils.ts
```diff
@@ -210,3 +211,49 @@ export function findMatchingTrainDialog(trainDialog: CLM.TrainDialog, trainDialo
     }
     return null
 }
+
+// Returns true is trainDialog1 is longer than trainDialog2
```
mattmazzola: typo, `if`

src/Utils/dialogUtils.ts
```diff
@@ -210,3 +211,49 @@ export function findMatchingTrainDialog(trainDialog: CLM.TrainDialog, trainDialo
...
+export function isLonger(trainDialog1: CLM.TrainDialog, trainDialog2: CLM.TrainDialog): boolean {
```
mattmazzola: The name is too generic. Looks like it should be comparing any two arrays and just so happens typescript can tell us.

mattmazzola:
@LarsLiden why did you merge this :(

The whole point of semantic versioning is that the versions mean something. We might as well just increment feature everytime like we used to if we do it this way

PR #1054, opened 2019-03-21
chore: ensure proper typing clones

mattmazzola:
There was many copies/clones of dialogs using `JSON.parse(JSON.stringify(...)` that dropped the types.
This puts them back.


mattmazzola:
Ok, changed.  Since the tests for merging still had the JSON conversion even though `deepCopy` was available I thought it was intentional not to use it. Anyways updated them all to use `deepCopy`

PR #1056, opened 2019-03-21
fix: enforce app name length on settings page

mattmazzola:
Combination of error on server not validating and not also checking on client allowed people to put arbitrarily long names for their app in the Settings page after they created it which had really bad efffect on our UI.

UI will now wrap long words, and prevent people from entering long names. There is also associated server PR to fix it there as well.

Before:
image

After:
image

Beyond immediate fix, but personally I would like to increase the limit. I think 30 is too short, maybe 50.
Also, our UI could better support long names like LUIS where it is dedicated bar / sub nav, but that's too much work

### DIFF
--- a/src/routes/Apps/App/Settings.tsx
+++ b/src/routes/Apps/App/Settings.tsx
@@ -229,8 +229,14 @@ class Settings extends React.Component<Props, ComponentState> {
     }
 
     onGetNameErrorMessage(value: string): string {
+        const MAX_NAME_LENGTH = 30
+
         if (value.length === 0) {
             return Util.formatMessageId(this.props.intl, FM.SETTINGS_FIELDERROR_REQUIREDVALUE)
+        }
+
+        if (value.length > MAX_NAME_LENGTH) {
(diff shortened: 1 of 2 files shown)


src/routes/Apps/App/Settings.tsx
```diff
@@ -229,10 +229,16 @@ class Settings extends React.Component<Props, ComponentState> {
...
 
     onGetNameErrorMessage(value: string): string {
+        const MAX_NAME_LENGTH = 30
+
         if (value.length === 0) {
             return Util.formatMessageId(this.props.intl, FM.SETTINGS_FIELDERROR_REQUIREDVALUE)
         }
 
+        if (value.length > MAX_NAME_LENGTH) {
+            return Util.formatMessageId(this.props.intl, FM.APPCREATOR_FIELDERROR_TOOLONG)
```
mattmazzola: Technically this is wrong since it's an `APPCREATOR` field, but in order to fix I'd have to change both settings and app creator to use generic / shareable names.

PR #1063, opened 2019-03-27
feat: add action placeholder in actions scorer

mattmazzola:
Add placeholder when the actions list is empty in a train dialog. Similar to the other lists on entities, actions, dialogs page etc.

The case would only occur if someone creates new train dialog with no actions, but that's what I usually do as I create new models by starting from train dialogs and creating actions as I go. 

We already have a + Actions button so it doesn't change much but I thought the user might feel lost and wonder if the list is broken since there are no items for any of the columns.

Before:
image

After:
image

Unrelated but as I was testing there is bug with Score Actions button when there is no actions.


mattmazzola:
Hmm, now this PR is stuck. Dang

PR #1065, opened 2019-03-27
feat: only show functions defined in callbacks

mattmazzola:
Depends on these PRs:

Only shows functions that were defined/given.  It will help users associate the each action in the list with the callback code for it better. 
Previously it showed both functions always which might confuse a user if they didn't actually define that function.

**Update: had wrong pictures before**

If we go further it could hide functions that don't define any custom arguments.
Issue here is that it violates easy rule to follow that if I wrote a function it will show up in this list.  Now it might be hidden if it didn't use arguments and confuse them thinking it was not defined.
image

This second view, only hides the logic function if it doesn't define custom arguments. This would make it still usable for the "set entity" case, and less confusing for people who do write render functions that just take things out of memory and don't use arguments:
image

Still think it's important to know what value is going in which argument so you know what you can/should do with that argument in code though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1796,9 +1796,9 @@
       }
     },
     "@conversationlearner/models": {
-      "version": "0.199.0",
-      "resolved": "https://registry.npmjs.org/@conversationlearner/models/-/models-0.199.0.tgz",
-      "integrity": "sha512-EH9qMwFkQCEgE5R3yARI6bOQcFYPBoeX3pgsDy7rsnQtfIuqznz2GXHV1lwx4/03WfABXbn6vriV969kMnmCgQ=="
+      "version": "0.199.1",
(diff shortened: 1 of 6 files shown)


mattmazzola:
Here is safest attempt which preserves the simple rule of if I defined it it will show up.
Forgot to include picture before

image

LarsLiden:
I still think that showing “memoryManager” and “result, memoryManager” is too noisy and makes it harder to read.

The first time you see it (for dev minded folks) it might be helpful, but after that it doesn’t add anything.

I’d suggest:

mattmazzola:
Ok, yea, I was also preferring the safe / simple option rather than removing functions with no arguments.  I think it's a higher risk to mislead the user by removing it and not explaining than to have extra information and them be aware of why things are happening.

PR #1066, opened 2019-03-28
chore: upgrade husky, typescript, and default package tag

mattmazzola:
Similar to:
https://github.com/Microsoft/ConversationLearner-SDK/pull/432
https://github.com/Microsoft/ConversationLearner-Models/pull/176

- Update husky, typescript, and ensure packages are published on `next` like SDK and Models


mattmazzola:
Hmm, seems to be Office Fabric problem with latest typescript. It will probably be fixed when upgrade to 6.x so can try then

PR #1068, opened 2019-03-29
build: split circleci

mattmazzola:
Right now Circle CI failure is misleading since it's doing two things but we only see one result. (normal building/test, running e2e tests)

This should be able to split them into separate workflows so we can put the required constraint back on the build and it should always pass, then move the less reliable e2e tests on separate workflow that may or may not fail


mattmazzola:
Hmm, seeing a bunch of "Cannot mkdir: Permission denied" errors

PR #1071, opened 2019-03-29
fix: bug in testing matching train dialog

LarsLiden:


src/Utils/dialogUtils.tsx
```diff
@@ -202,11 +202,14 @@ function doesExtractorStepMatch(extractorStep1: CLM.TrainExtractorStep, extracto
     // Only need to test the 1st Text Variation as they are equivalent w/in a round
     const labelEntities1 = extractorStep1.textVariations[0].labelEntities
     const labelEntities2 = extractorStep2.textVariations[0].labelEntities
-    if (labelEntities1.length !== labelEntities2.length) {
+
+    // Get unique ids
+    const entityIds1 = labelEntities1.map(le => le.entityId).filter((item, i, ar) => ar.indexOf(item) === i)
```
mattmazzola: Minor, but seems much more intuitive to use a `Set` which guarantees uniqueness instead of computing it using indices

PR #1072, opened 2019-03-29
feat: show user input array rather than first last in log list

LarsLiden:


src/Utils/dialogUtils.tsx
```diff
@@ -140,6 +140,19 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
     return best
 }
 
+export function logDialogSampleInput(trainDialog: CLM.LogDialog): string {
```
mattmazzola: The input is called `trainDialog` but the type `CLM.LogDialog` very confusing.

Think this would be simple change within the function instead of duplicating.
Only difference is how to get `userInput`

PR #1073, opened 2019-03-29
style: show user input array rather than first last in log list

LarsLiden:


mattmazzola:
You just merged another PR with same title:
https://github.com/Microsoft/ConversationLearner-UI/pull/1072
Doesn't make sense to be the same.

More like `fix: consolidate dialog sampling functions`

PR #1092, opened 2019-04-04
fix: bug 2003 label conflict dialog not showing up

LarsLiden:


src/routes/Apps/App/LogDialogs.tsx
```diff
@@ -888,12 +888,12 @@ class LogDialogs extends React.Component<Props, ComponentState> {
             definitions: null
         }
 
-        this.setState({
-            isEditDialogModalOpen: false,
-        })
-
         try {
             await this.props.createTrainDialogThunkAsync(this.props.app.appId, cleanedDialog)
+
```
mattmazzola: Ah there it is. I had been trying to move the line below `this.onCloseEditDialogModal()` inside the try catch and it wasn't working.  I think that also should only be run if it's successful.

LarsLiden: The two return statements will prevent the last call from happening it with a matched train dialog was found or there were conflicts

PR #1093, opened 2019-04-04
fix: bug 2000 settings value flashes on save

LarsLiden:
to merge after next release


src/routes/Apps/App/Settings.tsx
```diff
@@ -214,7 +214,7 @@ class Settings extends React.Component<Props, ComponentState> {
             trainingStatus: TrainingStatusCode.Completed,
             datetime: new Date()
         }
-        this.props.editApplicationThunkAsync(modifiedApp)
+        await this.props.editApplicationThunkAsync(modifiedApp)
         this.setState({
```
mattmazzola: Shouldn't this next update also be using the `modifiedApp` object instead of old `app` ?
`
Maybe this only worked because redux is updating the store which callls `componentDidUpdate` -> `updateAppState`?

LarsLiden: Yea, exactly.  Before the await it was updating from the old version.

mattmazzola: So you will remove since it's not needed?

LarsLiden: ah. good point

PR #1094, opened 2019-04-04
fix: bug 1992 tags input misaligned

LarsLiden:
merge after next release


mattmazzola:
Hmm, I don't think this one is needed after the recent fix to tags expanding beyond container

PR #1101, opened 2019-04-05
fix: bug 2014 changing of endsession action should be allowed

LarsLiden:


src/components/modals/ActionScorer.tsx
```diff
@@ -283,16 +285,21 @@ class ActionScorer extends React.Component<Props, ComponentState> {
         this.showAlreadySelectedPopUp = this.showAlreadySelectedPopUp.bind(this);
         this.onCloseAlreadySelectedPopUp = this.onCloseAlreadySelectedPopUp.bind(this);
     }
-    componentWillReceiveProps(newProps: Props) {
-        if (this.props.scoreResponse !== newProps.scoreResponse) {
+
+    componentDidUpdate(prevProps: Props) {
```
mattmazzola: Was there some timing condition that required it to be executed during the update instead of WillReceive time?
It seems like the move to be state was only optimization instead of calling `getScoredItems` every time?

LarsLiden: 1) scored items is needed by the column renderer so it needed to be state (plus more efficient)
2) changed to componentDidUpdate as componentWillReceiveProps is legacy (we should probably slowly move everything away from that)

PR #1103, opened 2019-04-06
Create End Session Model

SDESkowronski:
When I created the EndSession test case I manually created the model that I imported for that test thinking that I did not need to use automation to build it. However, I hit 2 significant bugs at the time which reproduced for me in that short time frame I was working on it (over the course of 30 minutes or so). I filed the bugs and in the morning nobody, not even me, could reproduce the bugs.

So I am adding in this test case thinking if it is some confluence of timing issues with services that those bugs will eventually occur again and we can catch it.


SDESkowronski:
I also added a test to detect a regression of "Bug 2026: EndSession Action causes description to be removed and second Train Dialog to be created."

cypress/integration/CreateModels/EndSession.spec.js
```diff
@@ -0,0 +1,102 @@
...
+  })
+
+  context('1st Train Dialog', () => {
+    it('Should create a new Train Dialog', () => {
+      modelPage.NavigateToTrainDialogs()
+      cy.WaitForTrainingStatusCompleted()
+      train.CreateNewTrainDialog()
+    })
+
+    it('Should train model to respond to "Hi"', () => {
```
mattmazzola: Only chose this as example but "'Should train model to respond to "Hi"'" don't seem like tests. Just steps to setup the larger test which is to verify behavior of EndSession actions.

You can't tell what this is testing by reading it.  As I've commented on previous PR I think the text should say what it's asserting.

cypress/integration/CreateModels/EndSession.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      train.TypeYourMessage('Hi')
+      editDialogModal.ClickScoreActionsButton()
+      train.SelectAction('Hello')
+    })
+
+    it('Should train model to respond to "Yo"', () => {
+      train.TypeYourMessage('Yo')
```
mattmazzola: Again on chose this as example but the pattern is in all the tests. It uses different helpers `train` then `editDialogModal` then `train`. The writer would expect to use the same "helper" for the same modal or part of the app.  This doesn't seem very intuitive.  I think people would only do this since the methods they want to use don't exist on the other object and they have to. 

Eventually I would like to see these reorganized so there isn't a switch or a better explanation as to why their needs to be a switch. I would expect if `train` is for operating on single train dialog it could also be used for editing.

cypress/integration/CreateModels/EndSession.spec.js
SDESkowronski: True, let me see what can do.

cypress/integration/CreateModels/EndSession.spec.js
SDESkowronski: Yes, this actually bothers me as well. Originally I set it up this way to follow the pattern set up by the product which had Train and Edit as distinct code files, but I agree that we can refactor this into something that makes more sense from the test code perspective. Adding it to my todo list and I'll keep it in mind on how best to handle this change. So for this PR we will leave it alone.

SDESkowronski:
Bugs 1968 & 1969 is what motivated me to create this test.

cypress/integration/CreateModels/EndSession.spec.js
SDESkowronski: So yes it is true that the wording is not so much test verification oriented, even though what it says is what is verified, it is more setup oriented because that is what it is doing. The wording as a whole, along with all the other "it()" calls makes sense, it accurately describes what is going on.

I really am not sure of how to make it better. I tried, put in some effort, but it only made it worse in my opinion. Also I really don't believe in that philosophy of yours, it is too restrictive. I don't believe it can be followed 100%. So I think this is something we should work out in our test meeting.

PR #1108, opened 2019-04-08
chore: remove eager client training status update and rely on server

mattmazzola:
Client tried to give user more immediate results by setting it to queued immediately so the user could have visual feedback instead of waiting for the first request to return which could take up to a second or so but this means the training status switches to queued when it might not be necessary such as only changing tags or description and re-saving.

The only I can see to fix this is to diff the rounds between the train dialogs on the client also to know if we should skip but this diffing and adding exceptions to the logic just makes it more complicated.

We can either we keep behavior as is to preserve fast change and deal with unnecessary Queued state, or we live with the delay and have it always be correct.

Important change is only one line:
https://github.com/Microsoft/ConversationLearner-UI/pull/1108/files#diff-c14f74716592e7bb12c99b6acfe97cebL34

Other stuff is minor


src/routes/Apps/App/Settings.tsx
```diff
@@ -211,8 +211,7 @@ class Settings extends React.Component<Props, ComponentState> {
             // packageVersions: DON'T SEND
             // devPackageId: DON'T SEND
             trainingFailureMessage: null,
-            trainingStatus: TrainingStatusCode.Completed,
-            datetime: new Date()
+            trainingStatus: TrainingStatusCode.Completed
```
mattmazzola: I had originally though this `datetime` would only be for the last training update, but this was also updated for settings.

Did we want to keep that so when you change settings it updates this time?

src/reducers/appsReducer.ts
```diff
@@ -31,7 +31,6 @@ const appsReducer: Reducer<AppsState> = produce((state: AppsState, action: Actio
                 return
             }
 
-            app.trainingStatus = TrainingStatusCode.Queued
```
mattmazzola: The actual change

src/routes/Apps/App/Settings.tsx
mattmazzola: Brings up my related question from Teams, it seems we added `lastModifiedDateTime` after `datetime` was here and now they seem to be updated inconsistently and it's not clear which is for what.

Since `datetime` is updated from Settings page update it is kind of a `lastModifiedDateTime`

The extra date time seems to make this confusing.

PR #1109, opened 2019-04-08
 fix: consolidate dialog editing logic

mattmazzola:
Consolidates dialog edit logic
- onInsertAction
- onInsertInput
- onUpdateHistory
- onChangeAction
- onChangeExtraction
- onDeleteTurn
- ononReplayTrainDialog

I'll have to go through the differences but mostly log dialogs were using specific forced editTypes while train dialogs were passing it through.

There was another case where `onUpdateHistory` differed and I'm not sure if it's a mistake or intended:
LogDialogs use `this.state.history`:
https://github.com/Microsoft/ConversationLearner-UI/blob/d75917bb5010abf4e89b0794def93a6e76b92d45/src/routes/Apps/App/LogDialogs.tsx#L756

TrainDialogs used `teachWithHistory.history`:
https://github.com/Microsoft/ConversationLearner-UI/blob/1c9a5c6a4cff6ea137f056c9bbec5ac2f6e11bb7/src/routes/Apps/App/TrainDialogs.tsx#L901

There were some differences between log and train I had question about though. 

For example tags are saved on TrainDialogs here:
https://github.com/Microsoft/ConversationLearner-UI/blob/1c9a5c6a4cff6ea137f056c9bbec5ac2f6e11bb7/src/routes/Apps/App/TrainDialogs.tsx#L871

But not for log dialogs here:
https://github.com/Microsoft/ConversationLearner-UI/blob/d75917bb5010abf4e89b0794def93a6e76b92d45/src/routes/Apps/App/LogDialogs.tsx#L729

It seems the DialogUtils is more than utils now and has some components inside it so I wanted to keep a file with regular typescript for the Editing


mattmazzola:
Do you know what the differences should be so I know which ones are accidental and which ones are intentional?

Seems related to discussion at lunch the other day about expansion of code paths and not being able to remember them all.  I could imagine there were previous bug fixes for train dialog editing that just never got carried over to log dialogs because they were missed

LarsLiden:
I think other than walking through them one by one, there’s not good way to tell.  My plan was to do a side-by-side comparison

mattmazzola:
Ok these are some of the differences I can see:

`onInsertAction` and `onInsertInput`
TrainDialogs - onUpdateHistory - selectionType, this.state.editType
LogDialogs - onUpdateHistory - SelectionType.NEXT, EditDialogType.LOG_EDITED

`onChangeAction`, `onChangeExtraction`
TrainDIalogs - onUpdateHistory - SelectionType.NONE, this.state.editType
LogDialogs - onUpdateHistory - SelectionType.NONE, EditDialogType.LOG_EDITED

`onEditTeach`
TrainDialogs - sets channel data
LogDialogs - does nothing
Merge - (always set channel data)

LarsLiden:
The first ones seem legit differences.  I think not setting channel data on LogDialogs was a miss, but would need to test

mattmazzola:
Also the changes in the description too, using `this.state.history` vs `teachWithHistory.history` or why we don't save `tags` and `description` for both

PR #1113, opened 2019-04-10
chore: show bot responses on default description

mattmazzola:
Shows bot responses in between user input so the sequence is more logical and linear.

Need to decide if this is better or not. Could argue that the bot responses are known to the trainers and don't as much value but I think it helps them make sense of the dialog when it's more complicated and they can't assume all the transitions in their head.

See: https://dev.azure.com/dialearn/BLIS/_queries/edit/2005/?triage=true

Currently image only shows text actions, but we could show the name of the code or card callbacks as well.

Before:
image

After:
image

### DIFF
--- a/src/Utils/dialogUtils.tsx
+++ b/src/Utils/dialogUtils.tsx
@@ -49,7 +49,7 @@ export function internalConflict(textVariation: CLM.TextVariation, trainDialog: 
 }
 
 export function activityIndexFromRound(trainDialog: CLM.TrainDialog, roundIndex: number | null, scoreIndex: number | null): number | undefined {
-    if (!roundIndex) { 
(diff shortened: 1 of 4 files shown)


LarsLiden:
Don’t think this one is a good idea.   The Bot responses don’t provide valuable information to distinguish the different train dialogs as they are all canned.  Given the limited space, having as many user input as possible is much better

mattmazzola:
Yes, I assumed that's why it was designed to only include user inputs but it seems confusing as is and wanted to see what other options we had to compare.

swadheen:
Agree with @LarsLiden  here.. User inputs are enough to give a view of the flow - and bot responses are known to the author - and are repetitive

PR #1114, opened 2019-04-10
build: enable 2.1 and pipelines to reduce redundant

mattmazzola:
In order to enable the ability to cancel the redundant builds we needed to be on 2.1


mattmazzola:
Should be able to auto cancel redundant builds now.  Meaning if we push a change to PR the tests for the latest commit don't wait for irrelevant/stale tests from previous commit.

image

PR #1115, opened 2019-04-10
fix: bug 2028 give warning when multi-tag non-multivalue

LarsLiden:
New warning message:

image


src/components/ExtractorResponseEditor/ExtractorResponseEditor.css
```diff
@@ -28,10 +28,15 @@
 .entity-labeler__custom-editor--read-only {
     background-color: rgba(0,0,0,0.025);
 }
+
 .entity-labeler__custom-editor--error {
     border-color: red;
```
mattmazzola: Hmm, not related to PR but I noticed this when you were demoing.  This should be the `var(--color-error)` from Office fabric to match our other errors.

src/components/modals/EntityExtractor.tsx
```diff
@@ -164,6 +165,29 @@ class EntityExtractor extends React.Component<Props, ComponentState> {
...
+
+    // Return list of non-multivalue entity names that have been labelled more than once
+    duplicateEntityNames(extractResponse: CLM.ExtractResponse): string[] {
+        const extractEntities = this.withoutPreBuilts(extractResponse.predictedEntities)
+
+        // Get list of entity ids that are tagged more than once
+        const multiEntityIds = extractEntities.map(pe => pe.entityId)
+            .filter(entityId => {
+                const count = extractEntities.filter(pe => pe.entityId === entityId)
+                return count && count.length > 1
```
mattmazzola: Wasn't sure why `count && ` was needed.  `filter` should always return arrays so this seems unnecessary.

PR #1116, opened 2019-04-10
Refactor Train & Edit Dialog - Test Grid Updated

SDESkowronski:
Updated the Test Grid - Should now contain most test scenarios (unless I missed something)

Refactored Train.BookMeAFlight test case into the new pattern so as to help me understand what it covered while updating the Test Grid.

Moved the code that was in EditDialogModal.js into Train.js since it was confusing while using the two for coding up test cases.


cypress/support/components/TrainDialogsGrid.js
```diff
@@ -33,3 +33,23 @@ export function GetDescription() { return helpers.StringArrayFromElementText('[d
 export function VerifyErrorIconForTrainGridRow(rowIndex) { cy.Get(`div.ms-List-cell[data-list-index="${rowIndex}"]`).find('[data-testid="train-dialogs-validity-indicator"]') }
 
 export function VerifyDescriptionForRow(row, description) { cy.Get(`div[data-item-index=${row}][data-automationid="DetailsRow"]`).find('span[data-testid="train-dialogs-description"]').contains(description) }
+
```
SDESkowronski: The code below was moved here from Train.js since it always should have been in this file.

SDESkowronski:
In another PR I also plan on moving and renaming "Train.js" to "TrainDialog.js" and moving it from the `support` folder into the `support\components folder`...but am waiting so that it does not totally make code reviewing impossible.

PR #1117, opened 2019-04-10
 fix: bug 2045 tags and description not preserved when editing a train dialog

LarsLiden:


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -709,7 +708,13 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         const editType = (this.state.editType !== EditDialogType.NEW && this.state.editType !== EditDialogType.BRANCH) ?
             EditDialogType.TRAIN_EDITED : this.state.editType
 
-        // TODO: Why does LogDialogs add tags but TrainDialogs do not?
+        // Update currentTrainDialog with tags and description
```
mattmazzola: Ah so they should be the same.  Stuff like this seems like it should be enforced in code otherwise it will just diverge again and keeping having regressions.

LarsLiden: these two could probably be moved to DialogEditing.   Think the only diff now is the editType which we could easily account for

PR #1120, opened 2019-04-11
fix: bug 1982 can open multiple entity pickers

LarsLiden:
Old behavior:
pickerold

New behavior:
pickernew


src/components/modals/EntityExtractor.tsx
```diff
@@ -58,7 +60,8 @@ class EntityExtractor extends React.Component<Props, ComponentState> {
             savedRoundIndex: 0,
             textVariationValue: '',
             newTextVariations: [],
-            entityTypeFilter: CLM.EntityType.LUIS
+            entityTypeFilter: CLM.EntityType.LUIS,
+            openPickerText: null
```
mattmazzola: Don't think the name of this is clear.  Is there a better way to describe it's use?

Initially I would think it sets the text of the open picker, but it seems to be actually just a copy of the latest extraction text for comparison.

LarsLiden: "activePickerText"?

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
```diff
@@ -322,10 +325,15 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
         if (pickerProps) {
             this.setState({
                 isSelectionOverlappingOtherEntities: pickerProps.isOverlappingOtherEntities,
-                isMenuVisible: pickerProps.isVisible,
                 menuPosition: pickerProps.position,
                 builtInTypeFilter: pickerProps.builtInTypeFilter
             })
+            if (this.props.isPickerVisible && !pickerProps.isVisible) {
```
mattmazzola: If both the `isPickerVisible ` and `onClosePicker()` came from props can the parent that set ` isPickerVisible` also call `onClosePicker()` since it laready know that information?  It seems like we're telling a child component to call a function instead of just calling the function.

src/components/modals/EntityExtractor.tsx
mattmazzola: I don't think that helps the situation. "open" and "active" are essentially the same thing.  The picker has it's own text and the extraction has text.

My understanding is that you only want to leave the picker open for the extractions that have the matching text. All those that don't match should be closed.

Would prefer
`openExtractionText`

PR #1122, opened 2019-04-12
fix: bug 588 flicker when closing chat window

LarsLiden:


LarsLiden:
Yeah, it isn’t complete clear to me what the path is but I think your intuition is correct.

mattmazzola:
But my intuition didn't give me an answer. This works and we don't know why... Things like this risky.

PR #1123, opened 2019-04-12
test: entity picker refactor and tests

mattmazzola:
Should give list of test cases:

```
  EntityPicker
    new model
      √ should open with 0 entities in the list (10571ms)
      √ should auto labeled the newly created entity (1169ms)
      √ should open with previously created entity in options (1143ms)
      √ should still list entity even if had been used and removed (539ms)
    existing model
      controls
        √ when pressing tab it should select the highlighted entity (8022ms)
        √ when pressing enter it should select the highlighted entity (424ms)
        highlight
          √ should list all the entities in the model (110ms)
          √ should open with the first item highlighted (133ms)
          √ when closing and reopening the picker after adjusting text and highlight should reset to empty and be at the top (385ms)
          arrow keys
            √ when pressing down arrow it should move the highlight down (125ms)
            √ when pressing up arrow it should move the highlight up (164ms)
            √ when pressing up arrow and the top it should rotate to the bottom (70ms)
            √ when pressing down arrow and the end it should rotate to the top (65ms)
      searching
        √ given a search input of X should highlighted characters X in the word (3974ms)
        √ should filter the list of entities to those matching search term (40ms)
```

Link to Cypress video:
https://microsoft.sharepoint.com/:v:/t/convsysresearch/EWFwI8IE0-9IqUlE4fkuVVIBUws9LD88gW9e0b7ganBjRA?e=hGwv1E

No training so it can runs all these tests in 35 seconds! Imagine if all these test cases were split across unrelated scenarios and in order to tell if something works you have to run all of them and wait for all of them.

Let me know if you think there are other cases.

Almost more important than the tests is the implementation changes to EntityPicker.
- Gets rid of the wrapping container using hooks as outlines in previous PR
- Allows you to think about effects of component due to changes of variables instead of thinking about component lifecycle like mount, didChange, unmound, etc.  Instead of intermixing unrelated concerns across the lifecycle methods, you get to keep them together so there's only one place to look.

In more practical terms, instead of:

```
componentDidMount() {
  doA()
  doB()
  doC()
}

componentDidChange() {
  if(prevB !== nextB) {
    doB()
  }

  doA()
}

componentWillUnmound() {
  doA()
  doC()
}

render() { }
```
It could be:
```
() => {
useA()
useB()
useC()
...
}
```


src/components/ExtractorResponseEditor/EntityPickerContainer.tsx
```diff
@@ -196,24 +195,23 @@ export default class EntityPickerContainer extends React.Component<Props, State>
     }
 
     render() {
```
LarsLiden: huh?

mattmazzola: Oops, was using it as reference to preserve functionality and forgot to delete it.

PR #1125, opened 2019-04-15
fix: bug 2058 bad replacement call when not needed

LarsLiden:
sourceTrainDialogId shouldn't be set when dialog is new

### DIFF
--- a/src/routes/Apps/App/TrainDialogs.tsx
+++ b/src/routes/Apps/App/TrainDialogs.tsx
@@ -345,6 +345,16 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         }
     }
 
+    // If editing an existing train dialog, return its Id, otherwise null
+    sourceTrainDialogId(): string | null {
+        if (this.state.currentTrainDialog 
+            && this.state.editType !== EditDialogType.BRANCH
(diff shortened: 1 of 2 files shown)


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -345,6 +345,16 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         }
     }
 
+    // If editing an existing train dialog, return its Id, otherwise null
+    sourceTrainDialogId(): string | null {
```
mattmazzola: Would this also be used in LogDialogs here?
https://github.com/Microsoft/ConversationLearner-UI/blob/master/src/routes/Apps/App/LogDialogs.tsx#L636

Seems like it is not specific to train dialogs

LarsLiden: Log Dialogs should never be in the "NEW" state as they always start with an existing log.  In fact they should never be in "BRANCH" either so that part of the check can go away

LarsLiden: Was able to simplify this quite a bit

PR #1128, opened 2019-04-17
fix: bug 2059 internal label confict errors

LarsLiden:
Fix internal label conflict errors.  For some instances requires that the user save a partial edit dialog before continuing to edit. Eventually the server needs to not check for label conflicts for edited train dialogs when creating a teach session with context, but it currently does to that so we use a save pop-up as a workaround
image


src/services/client.ts
```diff
@@ -569,15 +569,21 @@ export default class ClClient {
     }
 
     //AT.CREATE_TEACH_SESSION_FROMHISTORYASYNC
-    async teachSessionFromHistory(appId: string, trainDialog: CLM.TrainDialog, userInput: CLM.UserInput | null, userName: string, userId: string): Promise<CLM.TeachWithHistory> {
+    // filteredDialog = dialog to ignore when checking for conflicting labels
+    async teachSessionFromHistory(appId: string, trainDialog: CLM.TrainDialog, userInput: CLM.UserInput | null, userName: string, userId: string, filteredDialog: string | null): Promise<CLM.TeachWithHistory> {
+        let url = `${this.baseUrl}/app/${appId}/teachwithhistory?username=${userName}&userid=${userId}`
+        if (filteredDialog) {
```
mattmazzola: This is misleading, it's a dialog id and not a dialog. Would prefer: `filteredDialogId`

It's a bigger change and probably not for this PR but would prefer the name to specify how it's filtering. It doesn't indicate if the filters is only id's given or excluding the id's given. Would make more sense to name is `excludeDialogIds`  so you don't have to remmeber which API's are positive or negative conditions

src/components/modals/EditDialogModal.tsx
```diff
@@ -237,6 +247,24 @@ class EditDialogModal extends React.Component<Props, ComponentState> {
...
+            isSaveConflictModalOpen: false
+        })
+    }
+
+    @OF.autobind
+    onSaveConflictCancel() {
+        // Increment webchat key to reset and clear last input
+        this.setState({
+            isSaveConflictModalOpen: false,
+            webchatKey: this.state.webchatKey + 1
```
mattmazzola: Interested why changing the key removes the last input.

src/components/modals/EditDialogModal.tsx
```diff
@@ -303,11 +331,27 @@ class EditDialogModal extends React.Component<Props, ComponentState> {
...
+                    return
+                }
                 await this.props.onContinueDialog(newTrainDialog, userInput)
             }
         }
     }
 
+    // TEMP: until server can exclude label conflicts with self, we need
+    // to check for them and force save before we can add a turn 
+    showInternalLabelConflict(): boolean {
```
mattmazzola: Minor but the naming on this is misleading: `showInternalLabelConflict` doesn't always show the conflict.  Since it's temporary I guess it won't be in the code long but that's what they always say.

The methods that consume or use this treat it like another `hasInternalLabelConflict` which is clear about returning true or false and don't care if it opens the modal or not they just return early to prevent execution

There is similar code:
```
            if (await this.hasConflicts(textVariations)) {
                return
            }
```

Not sure about new name though if you call it `hasInternalLabelConflict` to wrap the DialogUtil method it doesn't indicate it will open the modal

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -828,14 +836,12 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
...
-            const originalId = this.state.currentTrainDialog
-                ? this.state.currentTrainDialog.trainDialogId
-                : null
+
             this.setState({
                 history: teachWithHistory.history,
                 lastAction: teachWithHistory.lastAction,
                 currentTrainDialog: trainDialog,
-                originalTrainDialogId: originalId,
+                originalTrainDialog: this.state.currentTrainDialog || this.state.currentTrainDialog,
```
mattmazzola: I'm guessing one of these supposed to be `this.state.originalTrainDialog` ? 

Otherwise remove one of them

src/services/client.ts
LarsLiden: API calls it "filterDialog" so I went with that but I agree I prefer if it says what it is

src/components/modals/EditDialogModal.tsx
LarsLiden: It forces a redraw of webchat from the the train dialog (which hasn't yet been updated)

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: fixed

src/components/modals/EditDialogModal.tsx
mattmazzola: Yea, I know changing key forces remount but I didn't understand how the item got there is we didn't put it there.

I didn't think we gave the web chat a dialog only gave it `history` which was a list of activities. Is that correct?
So it has prop of `history` and internal state of activities and by forcing refresh it will re-use activities from the `history` prop but reset internal state?

LarsLiden: When the user types in a new input, the webchat display updates to show the new input, but the train dialog doesn't update as we detect the conflict.  The history doesn't have the new input

PR #1138, opened 2019-04-22
Artifact Test Recordings and Rotate LUIS Auth Keys

SDESkowronski:
1) Saves our test recordings to the CircleCI Artifacts Folder
2) Rotates LUIS Authoring Keys rather than randomly assigning them.


SDESkowronski:
Constraints are not clearly stated from what I could find. Here is what I did find...
[https://circleci.com/docs/2.0/artifacts/#artifacts-overview]

> Artifacts are stored on Amazon S3 and are protected with your CircleCI account for private projects. There is a 3GB curl file size limit. Artifacts are designed to be useful around the time of the build. It is best practice not to rely on artifacts as a software distribution mechanism with long term future guarantees.

PR #1140, opened 2019-04-24
Spec File Rename

SDESkowronski:
- Repathed the root folder that the Regression Tests are in
- Runs tests based on Glob pattern for the new Regression Test folder
- Renamed some of the Spec Files
- Fixed a bug


.circleci/config.yml
```diff
@@ -78,7 +78,7 @@ jobs:
...
+      - run: npm run cypress -- run --record false --spec  "cypress/integration/Regression/**/*.spec.js"
```
mattmazzola: Good to see this change.

What are you using to classify or qualify tests as necessary to be in this folder or not in it?
I think any product change that breaks a test is a regression so I didn't quite understand the need for specific "regression" folder. Are the tests not in it are considered extraneous or rare edge cases unnecessary to spend time testing?

I was expecting to see something more traditional like smoke tests to catch major things then all remaining to catch any others. This way tests still fail early but still have coverage.

I will probably move some of my tests in there as it's also important to know they still pass.

SDESkowronski: At this point the only tests excluded are your, until you put them in there, and those that are more like tools that I use for specific debugging purposes, or loop tests that are meant to catch rare errors.

The point of the folder is to make GLOB identification of the tests we run easier.

Yes, at some point we should consider breaking our tests out into groups like SMOKE or BVT tests, but for now it feels like the relatively small number of tests we have serve both purposes of SMOKE and REGRESSION.

PR #1141, opened 2019-04-26
Fix Delete All Test Models test case

SDESkowronski:
We use a Build Key that recycles every 26 builds, which gives us an effective 13 unique keys before they are reused. This is plenty to ensure we don't reuse the same key.

We delete based on the key, if the model name has the Build Key in it that matches the current build we always delete it. If not we delete the model if the created time is 5 minutes back or longer.


SDESkowronski:
The code in zTemp.spec.js is code to test the algorithm that determines if a model is eligible for deletion. I'm keeping that around for a few weeks in case I find a bug and need to retest. There is also added logging that I will eventually delete once I feel confident this works as I planned.

SDESkowronski:
We use a Build Key that recycles every 26 builds, which gives us an effective 13 unique keys before they are reused. This is plenty to ensure we don't reuse the same key.

We delete based on the key, if the model name has the Build Key in it that matches the current build we always delete it. If not we delete the model if the created time is 5 minutes back or longer.

PR #1142, opened 2019-04-26
feat: add set entity actions

mattmazzola:
Pending PR's for server, models, and SDK

- Adds `SET_ENTITY` action type
- Enables `ENUM` entities
- Adds warnings explaining `SET_ENTITY` actions are immutable 
- Prevents deletion of enum entities or values when used within associated SET_ENTITY actions. (Most restrictive for now.  Plan to relax constraints later but have to deal with removal of actions and invalidation of train dialogs which has large fallout and could be worse for user.)
- Adds ability to automatically populate list of action scorer with possible SET_ENTITY action and automatically create the action if selected
(There are lot of typescript casting to convert from ScoredAction to Action but that's pending schema refactor)

image

There was TODO about making the `handleActionSelection` function take in an action, we currently do it for SET_ENTITY actions since it was required, but left the other actions as if to minimize changes. It's probably good time to take that task on soon to simplify code.

Tests - Let me know of other cases if you think of them.
In another PR I will look at allowing inline enum entity creation which would be other test cases but didn't want to introduce that feature yet.
```
  Set Entity Actions
    model behavior
      enum entity deletion
        √ should prevent deletion of enum entities used within set entity actions (22792ms)
        √ should prevent deletion of enum VALUES used within set entity actions (5039ms)
        √ should allow deletion of unused enum values (4725ms)
        √ should allow deletion of enum entities NOT used within actions (6938ms)
      manual creation
        √ should show correct form state when SET_ENTITY type is selected (10998ms)
        √ should reset the value of the fields when changing the action type (6385ms)
        √ should not allow creating the action unless entity and enum are set (5255ms)
        √ should prevent creating duplicate actions (same entity and enum value) (8004ms)
      editing
        √ should show the entity name and enum value in the fields (4820ms)
        √ should not allow saving changes to SET_ENTITY actions (239ms)
      automatic creation
        √ should create a placeholder action for each possible enum value (12736ms)
        √ when clicking the placeholder action it should create a default set entity action then take it (15783ms)
    scenario using set entity actions
      √ should successful place fast food order using context for instead of labels (82107ms)
```
Video:
https://microsoft.sharepoint.com/:v:/t/convsysresearch/EXjJrEIlenpOuHyOA2qqbNUBRszoaHRMfY_ponbdvEzP7g?e=zu6apV


src/Utils/util.ts
```diff
@@ -158,4 +159,72 @@ export const returnStringWhenError = (s: string) => {
...
+
+export const setEntityActionDisplay = (action: CLM.ActionBase, entities: CLM.EntityBase[]): [string, string] => {
+    let name = `MISSING ENTITY`
+    let value = `MISSING VALUE`
+
+    const entity = entities.find(e => e.entityId === action.entityId)
+    if (entity) {
+        name = entity.entityName
+        if (entity.entityType !== CLM.EntityType.ENUM) {
+            value = `Entity Is Not Enum!`
```
LarsLiden: Do we need to internationalize this?

src/components/ToolTips/ToolTips.tsx
```diff
@@ -179,6 +181,10 @@ export function getTip(tipType: string) {
             return render(FM.TOOLTIP_ACTION_CARD_TITLE, [FM.TOOLTIP_ACTION_CARD])
         case TipType.ACTION_END_SESSION:
             return render(FM.TOOLTIP_ACTION_END_SESSION_TITLE, [FM.TOOLTIP_ACTION_END_SESSION])
+        case TipType.ACTION_SET_ENTITY:
+            return render(FM.TOOLTIP_ACTION_SET_ENTITY_TITLE, [FM.TOOLTIP_ACTION_SET_ENTITY])
+        case TipType.ACTION_ENTITY_VALUE:
+            return render(FM.TOOLTIP_ACTION_SET_ENTITY_TITLE, [FM.TOOLTIP_ACTION_SET_ENTITY])
```
LarsLiden: Not sure what context these are used in, but these two different tooltips seem to be showing the save values?  Is this work TBD?

src/components/modals/ActionScorer.tsx
```diff
@@ -95,7 +95,7 @@ function getColumns(intl: InjectedIntl): IRenderableColumn[] {
                     return (
                         <OF.PrimaryButton
                             data-testid="action-scorer-button-clickable"
-                            onClick={() => component.handleActionSelection(action.actionId)}
+                            onClick={() => component.handleActionSelection(action.actionId, action)}
```
LarsLiden: Why not just pass action?  Of is there a case where the actionId is coming from a different action?

src/components/ToolTips/ToolTips.tsx
mattmazzola: Yea, the tool tips don't really add value.  It's trying to explain what an Entity dropdown is. I would think it's obvious it is a list of entities, but I still added it since we have pattern of doing that for all our fields.

I agree it should add more detail but I did think it's probably best to keep these fields (tooltips) the same. I thought a higher level explanation of what set entity actions do such as explaining how they're used for context sensitive situations when label value alone is ambiguous.

What do you think we should put in the panel?

LarsLiden:
Let’s play with it for a while – get some feedback from Michael to see what would be helpful.

src/components/modals/ActionScorer.tsx
mattmazzola: Yea, talked about that in the description:

> There was TODO about making the handleActionSelection function take in an action, we currently do it for SET_ENTITY actions since it was required, but left the other actions as if to minimize changes. It's probably good time to take that task on soon to simplify code.

I can't remember exactly but I think there where the id that it was given came form either ScoredAction or Action and this mean we couldn't always pass in the same thing and would hve to do the search for action anyways, and didn't want to risk the change yet but it's very confusing and would also like to simplify.

src/Utils/util.ts
mattmazzola: Yea, we should although I wanted to keep the Util simple and there many other places that use "MISSING ACTION" and I was following that pattern for "MISSING ENTITY"

Given that we made entities used in SET_ENTITY actions non-deletable this case should never happen but we had to be defensive against it for future.

I guess we can make item in backlog to do all of these.

PR #1143, opened 2019-04-29
test: partial conversion of tests to typescript

mattmazzola:
- Converts tests outside tests folder to typescript
- Consolidates selectors and constants
- Types some of the custom commands on `cy`

Seems the log conversion test is broken due to another change in entity conflict detection and need to see what's going on there.


cypress/plugins/index.js
```diff
@@ -11,10 +11,14 @@
...
 // the project's config changing)
 const path = require('path')
+const cypressTypeScriptPreprocessor = require('./cy-ts-preprocessor')
 
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+
+  on('file:preprocessor', cypressTypeScriptPreprocessor)
+
```
SDESkowronski: Where is it that TS preprocessor errors will show up, where will I see them if there are any?

PR #1148, opened 2019-05-01
Added test scenarios

SDESkowronski:
While reviewing code in order to update our test grid spread sheet, I found minor bugs to fix and opportunities to easily cover a few extra test scenarios.

Also the recent addition of the Merge Train Dialogs feature was causing many tests to not complete correctly, and with one of the other changes it actually broke the test. So I fixed the `train.Save` function so that it looks for that Merge modal popup and clicks the `Save As Is` button when it is there.


cypress/support/Train.js
```diff
@@ -351,7 +352,7 @@ export function VerifyChatTurnIsAnExactMatch(expectedTurnText, expectedTurnCount
 function VerifyChatTurnInternal(expectedTurnCount, turnIndex, doVerification) {
   cy.WaitForStableDOM()
   let chatMessages
-  cy.wrap(undefined).should(() => { 
+  cy.wrap(1).should(() => {
```
SDESkowronski: It turns out that using `undefined` here prevents Cypress from invoking their retry logic, which was the whole point of wrapping it like I did. I noticed that while debugging changes to the `SaveAsIsVerifyInGrid` function.

cypress/integration/Regression/CreateModels/AllEntityTypes.spec.js
```diff
@@ -41,9 +58,15 @@ describe('All Entity Types - CreateModels', () => {
       entities.CreateNewEntity({ type: 'Programmatic', name: 'programmaticMultiValuedEntity', multiValued: true})
     })
 
+    // Alternate testing of multiValued and negatable on different days 
+    // so that we test these in combination regularly.
```
mattmazzola: Why does it need to be random or based on day of year?

If it's only there to get more coverage just add more tests. The tests should be consistent and not matter which day you choose to run them. I'm imagining explaining to someone that they can only test that case on even days of the year and them being in disbelief.  This just adds complexity to think about.

I think last time I looked these tests take ~7 seconds to create each entity and total time was ~2 minutes, 25 seconds on a tests that are highly unlikely to fail and delay info about the more important tests. Another reason to split into smoke / remaining.

I suppose there are shortcuts like importing a model with all variations of entities which would implicitly test server and luis, and the UI test could only verify a few of the type and dropdown functionality but I'm hesitant to do that either.

cypress/integration/Regression/CreateModels/AllEntityTypes.spec.js
```diff
@@ -7,6 +7,23 @@ import * as models from '../../../support/Models'
 import * as entities from '../../../support/Entities'
 import * as helpers from '../../../support/Helpers'
 
+// Alternately return true or false.
+// Starts out with different value depending on the day of the year.
+// Can't do this with simple variables due to the way Cypress works.
```
mattmazzola: > Can't do this with simple variables due to the way Cypress works.

What did you mean by this?  Seems like a closure thing instead of Cypress thing.

cypress/integration/Regression/CreateModels/AllEntityTypes.spec.js
SDESkowronski: I agree with most of what you are saying here. I was trying to increase coverage while not impacting the test run time as much. But we have better ways of dealing with this so...

Since we are at the point of needing to break tests into Smoke vs Full/Regression Testing I'll be sure this test is in the full test list and fix it to test every possible combination when that change over is made.

cypress/integration/Regression/CreateModels/AllEntityTypes.spec.js
SDESkowronski: You are correct. I fixed the comment.

PR #1156, opened 2019-05-07
 feat: allow removing action from dialogs

mattmazzola:
Adds checkbox to the action delete modal to allow deleting from dialogs.
Resets or defaults to `false` every time the modal is opened so people have to opt-in to the usage.

image

https://dev.azure.com/dialearn/BLIS/_workitems/edit/1934

I think the `ConfirmCancelModal` is getting to the point of being over used as we keep adding parameters and more conditions.  E.g. Most of them don't need this option but yet it has be in all of them.  I would consider using a dedicated `ActionDeleteModal`.

Fixes ADO#1934


LarsLiden:
Wondering if "Delete Action from All Train Dialogs" might be clearer.
Alternatively, a radio button could be even clearer.  Something like
- Delete Action from Train Dialogs where it is employed
- Leave empty stub action in place of deleted action

mattmazzola:
I think you're right about using radio buttons here.

I will hold it for a later PR because it's causes many changes in other tests expecting confirm cancel modal and selectors.

```
  Action Deletion
    preserve action placeholder
      √ should mark the dialog using action as invalid (15652ms)
    remove scorer step
      terminal actions
        √ should mark dialogs invalid if action user other than only score step of last round (5379ms)
        √ should remain valid if action was the only scorer steps of last round (812ms)
        √ should become invalid if last round contains multiple scorer steps (non-terminal) (778ms)
        √ should become invalid if terminal action was used in a previous round and last round (778ms)
      non-terminal
        does not affect memory
          √ removing TEXT action from dialog should not modify validity (4438ms)
          - removing CARD action from dialog should not modify validity
        does affect memory
          √ removing SET_ENTITY action from dialog should require replay and show WARNING (4660ms)
          - removing API action from dialog should require replay and show WARNING
      preserve worst validity
        √ dialog should remain INVALID after deletion operation which is VALID (9798ms)
```

Video: https://microsoft.sharepoint.com/:v:/t/convsysresearch/EbGdwXR9aJpDiuBBxVQpvjEBvU5q24aIONsxwCtxFNRMxw?e=tXB9qM

PR #1159, opened 2019-05-17
Create API Actions

SDESkowronski:
General purpose code for creating API Actions and a new test case that is almost working except for a blocking issue in the UI.

See: [Bug 2132: Automation cannot trigger 2nd Entity picker in API Action arguments](https://dialearn.visualstudio.com/BLIS/_workitems/edit/2132)


cypress/integration/Regression/CreateModels/ApiCallbacks.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      })
+    })
+    
+    it('Should create RenderTheArgs Action', () => {
+      actions.CreateNewActionThenVerifyInGrid({ 
+        responseNameData: 'RenderTheArgs',
+        type: 'API',
+        logicArgs: ['$1stArg{enter}', '$2ndArg{enter}', '333', '4444', 'five', 'six', 'seven'],                                          
+        renderArgs: ['$1stArg{enter}', '$2ndArg{enter}', 'three', 'four', '55555', '666666', '7777777'],
+        validateApiResponse: 'RenderTheArgslogic(memoryManager, firstArg, secondArg, thirdArg, fourthArg, fifthArg, sixthArg, seventhArg)firstArg:"$1stArg"secondArg:"$2ndArg"thirdArg:"333"fourthArg:"4444"fifthArg:"five"sixthArg:"six"seventhArg:"seven"render(result, memoryManager, firstArg, secondArg, thirdArg, fourthArg, fifthArg, sixthArg, seventhArg)firstArg:"$1stArg"secondArg:"$2ndArg"thirdArg:"three"fourthArg:"four"fifthArg:"55555"sixthArg:"666666"seventhArg:"7777777"'
```
mattmazzola: I don't think you would want to validate the response this way.  The display/format has changed and might to change again which would break tests.

The string is also not an expected format for people writing tests.  This relies on the joining of text without newlines and other internal behavior. 

Many disagreements with the function but I guess we'll put that aside.  If you want to do this type of validation I try to find a way to construct this string from the other input (response and logic /render args) so is always correct and people don't have to manually type it. Could easily be a typo or wrong character.

cypress/integration/Regression/CreateModels/ApiCallbacks.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      })
+    })
+
+    it('Should create TextCard Action', () => {
```
mattmazzola: What is this testing?  What is a TextCard?

The test says create a TextCard action which is two different action types, but then inside the method it creates an API.

cypress/support/components/ActionsGrid.js
```diff
@@ -19,22 +19,38 @@ export function ValidateRequiredEntities(requiredEntitiesFromResponse, additiona
...
+]
 
 // IMPORTANT: Call this method before calling any of the Validate* methods.
-export function GetEndSessionRowToBeValidated(response) {
-  cy.Get('[data-testid="action-scorer-session-response"]')
-    .ExactMatch('EndSession')
-    .siblings('[data-testid="action-scorer-session-response-user"]')
-    .ExactMatch(response)
+export function GetRowToBeValidated(actionType, textId) {
+  let i = typeSelectorPairs.findIndex(typeSelectorPair => typeSelectorPair.type === actionType)
```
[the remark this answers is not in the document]
SDESkowronski: Thanks. I had something else I was trying in tandem to do that needed the index and later realized I did not need it, but then forgot to fix this. I've fixed it now.

cypress/support/components/ActionModal.js
```diff
@@ -26,6 +29,48 @@ export function TypeResponse(textToType) {
...
+  cy.Get('label')
+    .contains(apiArgLabel)
+    .siblings('div.editor-container')
+    .find('div[data-slate-editor="true"]')
+    .then(elements => {
+      if (elements.length != args.length) {
+        throw new Error(`Test Code Error: The API ${apiArgLabel} takes ${elements.length} arguments, but test code supplied ${args.length}`)
+      }
+
+      cy.DumpHtmlOnDomChange(true)
```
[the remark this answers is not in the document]
SDESkowronski: While this may be possible, I expect that it will be complex and take more time than I want it to. This code executes in the background inside of MonitorDocumentChanges.js, so I don't want it to slow the tests down. I turn it on when I need it and turn it off when I don't. It gets logged to the log file and I use external tools, BeyondCompare being the primary one, to compare the differences when I need to.

cypress/integration/Regression/CreateModels/ApiCallbacks.spec.js
SDESkowronski: Even if I construct from the arguments, if the format changes on the UI I'll still have to change the test code.

I don't have to worry about typos since I grab this string from the logs...I run the test with this value undefined, grab the string from the logs and past it into the code...of course I look it over for a sanity check before I do to ensure the output makes sense. So if it does fail because the UI changes, that change will appear in the log file...I just have to copy and paste it into the test code to fix the test error.

Also when output like this changes, it is nice for me to discover that and have to change the code. This way I am one more pair of eyes on the change to confirm it makes sense to someone besides the developers who put it there.

cypress/integration/Regression/CreateModels/ApiCallbacks.spec.js
SDESkowronski: TextCard is my name for the API Callback. It is just a card with a label and text inside of the box.

cypress/support/components/ActionsGrid.js
```diff
@@ -19,22 +19,38 @@ export function ValidateRequiredEntities(requiredEntitiesFromResponse, additiona
 // so the disqualifyingEntities parameter allows the caller to specify entities not found in expectedEntities.
 export function ValidateDisqualifyingEntities(expectedEntities, disqualifyingEntities) { ValidateEntities('[data-testid="action-details-disqualifying-entity"]', '[data-testid="action-details-empty-disqualifying-entities"]', expectedEntities, disqualifyingEntities) }
 
+// In order to get the 'validateResponse' parameter right, first run a test with this undefined,
+// then look in the log to see the actual value, then add it to the code.
+export function ValidateApi(validateApiResponse) {
+  cy.Get('@responseDetailsRow')
```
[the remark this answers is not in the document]
SDESkowronski: I agree this is a bit fragile. I should be able to wrap this up in a class so the coupling is automatic to mitigate the issue...should be in the next PR.

PR #1161, opened 2019-05-20
Error Handling Test Cases

SDESkowronski:
This fixes the issue with error handling. This PR is really more of an FYI, since I am going to push it into master immediately so I don't have to deal with broken tests any longer than nessary. However, if there is an issue with the changes I'll be happy to discuss and alter it if necessary.

### DIFF
--- a/cypress/integration/Regression/ErrorHandling/ActionUnavailable.spec.js
+++ b/cypress/integration/Regression/ErrorHandling/ActionUnavailable.spec.js
@@ -46,7 +46,7 @@ describe('Action Unavailable - ErrorHandling', () => {
 
     it('Should verify the specifics of the error', () => {
       train.SelectChatTurnStartsWith('Hello')
-      train.VerifyErrorMessage('Action is unavailable')
+      train.VerifyErrorMessage('Action is unavailable')
       train.VerifyChatTurnIsAnExactMatch('Hello [$name]', 2, 1)
     })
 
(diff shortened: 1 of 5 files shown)


cypress/integration/Regression/ErrorHandling/ActionUnavailable.spec.js
```diff
@@ -46,7 +46,7 @@ describe('Action Unavailable - ErrorHandling', () => {
 
     it('Should verify the specifics of the error', () => {
       train.SelectChatTurnStartsWith('Hello')
-      train.VerifyErrorMessage('Action is unavailable')
+      train.VerifyErrorMessage('Action is unavailable')
```
SDESkowronski: Had to add in this special character that we use as a button to bring up more information about the error message. This is due to not zooming in on the exact DOM element like we were previously doing.

mattmazzola: Hmm, yea, this seems related to what I mentioned in Teams. Would be more appropriate to test that an action unavailable error was displayed and not the actual text. Or using contains instead of exact so you don't have to change the test when we change the message.

Something to have a testid for each error case like:
```
data-testid="dialog-modal-error-action-unavailable"
data-testid="dialog-modal-error-action-missing"
data-testid="dialog-modal-error-action-logic"
data-testid="dialog-modal-error-action-render"
...more cases...
```

SDESkowronski: Actually, I worked it out. Don't need to include the special character.

The reasons I like to test the exact error message is:

1. When reviewing code it is easier to relate to what I see on screen if I use the exact error message.
2. If devs change the text, either accidently or intentionally, it will fail and I can review the change and make sure it is correct and makes sense...another eyes on messages to the user is usually helpful.

PR #1163, opened 2019-05-22
feat: use radio buttons for action deletion and add help panel information

mattmazzola:
Video: https://microsoft.sharepoint.com/:v:/t/convsysresearch/EfI74vzjYN5HjkAxel4x6KMBD-Cr7w13avC_pKo0zmEM7Q?e=CVATSb

Also changes it so the option to remove placeholders doesn't show up when the action is not used.

I assume this will break some of the existing tests relying on ConfirmCancel modal.

Wonder how this will change when we add the ability to replace actions as well.


cypress/integration/actionDeletion.spec.ts
```diff
@@ -101,13 +101,15 @@ describe('Action Deletion', () => {
                 cy.get(s.common.spinner, { timeout: constants.spinner.timeout })
                     .should('not.exist')
 
-                cy.get(s.confirmCancelModal.optionRemovePlaceholder)
+                cy.get(s.actionDeleteModal.deleteType)
+                    .find('label')
```
mattmazzola: I had tried to avoid using the HTML sub structure to get the dom element since it can change using various properties in OF and they either didn't work or just transferred complexity from tests to app and didn't improve it. In 
 this case the HTML is coming from Office fabric instead of our App so hopefully means it won't change as frequently.

For some reason using `:nth-child(2)` wasn't working, will have to investigate if I run into this again

mattmazzola: After looking more closely it seems the do allow any custom attributes such as `data-testid`s on the choices.  It wasn't documented on the website but they demo it in one of the codepens.

src/react-intl-messages.ts
```diff
@@ -616,6 +621,10 @@ export default {
         [FM.ACTIONSCORER_COLUMNS_ENTITIES]: 'Entities',
         [FM.ACTIONSCORER_COLUMNS_ISTERMINAL]: 'Wait',
         [FM.ACTIONSCORER_COLUMNS_TYPE]: 'Type',
+        
+        [FM.ACTION_DELETE_INUSE_TITLE]: 'Deleting actions used by train dialogs',
+        [FM.ACTION_DELETE_INUSE_PLACEHOLDER_PRESERVE]: 'This operation will preserves the reference to the action leaving a placeholder. All the dialogs using the action will become invalid.  This option can be useful to see where and how the action was used as you correct the dialogs.',
```
LarsLiden: Wondering if we might want to suggest that this option is useful if the intention is the Replace the action with another one, whereas the other option is for just removing the action?

mattmazzola: Yea, I'd meant for the "correct the dialogs" to be generic enough that you could fix it however you wanted, replace as one option but I added text at the end to say " or replace the action" so it's more clear.

I think we planned to have an actual replace action feature where you could select a source and target to do the swap, but that UI and feature still needs to be decided. This is interim solution to get ability to remove when you realize it was not needed and don't want to invalidate your dialogs.

PR #1166, opened 2019-05-29
Api Multiple Exceptions Test Automation

SDESkowronski:
- Added two new test suites:

1. ErrorHandling/ApiCreateMultipleExceptions
2. ErrorHandling/ApiVerifyMultipleExceptions

- Fixed bugs found in existing helper functions while debugging the new tests.
- Moved some verification code that was in ScorerModal.js into Train.js where it should have been in the first place.
- Extended the Train/ApiCallback test case with some edit verifications.


src/components/ExtractorResponseEditor/FuseMatch.tsx
```diff
@@ -14,5 +14,5 @@ interface Props {
  * Display Fuse.io search match with characters from search input having custom style such as highlight or bold
  */
 export default function ({ matches }: Props) {
-    return <span>{matches.map((m, i) => <span className={`match-string ${m.matched ? 'match-string--matched' : ''}`} key={i}>{m.text}</span>)}</span>
+    return <span>{matches.map((m, i) => <span data-testid="entity-picker-match-option" className={`match-string${m.matched ? ' match-string--matched' : ''}`} key={i}>{m.text}</span>)}</span>
```
mattmazzola: This isn't a correct `data-testid` as it's not guaranteed to be an `entity-picker` related match.  This is used in many places. Should put data-testids for entity picker in the entity picker component.

Would either move this testid up for use a selector that first specifies entity picker then gets this option within.

The only think you know is this is a fuse match so it could be `data-testid="fuse-match-option"`

PR #1175, opened 2019-06-07
feat: add new prebuilts (geographyV2, personName, keyPhrase)

mattmazzola:
As discussed on Teams, update to keep closer parity with LUIS capabilities

### DIFF
--- a/src/types/PreBuiltEntities.ts
+++ b/src/types/PreBuiltEntities.ts
@@ -13,8 +13,12 @@ export const PreBuilts = {
     Age: "age",
     URL: "url",
     Email: "email",
-    Phone_number: "phonenumber"
+    Phone_number: "phonenumber",
+    Key_Phrase: "keyPhrase",
+    Person_Name: "personName",
+    GeographyV2: "geographyV2",
 }
+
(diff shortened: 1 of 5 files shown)


src/types/PreBuiltEntities.ts
```diff
@@ -13,8 +13,12 @@ export const PreBuilts = {
     Age: "age",
     URL: "url",
     Email: "email",
-    Phone_number: "phonenumber"
+    Phone_number: "phonenumber",
```
LarsLiden: Is the inconsistent casing coming from LUIS?

mattmazzola: Idk, it was that way on the server too. I just left it the way it was as it risks changing more things but I can change.

In my opnion if you use `_` underscores the other characters should be consistent casing and the underscores delimit the words. If you use casing do note seperation of words then you don't use `_` at all.

E.g.
```
personName
PersonName
PERSON_NAME
person_name 
Person_name - Weird mix of conventions
```
Since this is an `enum` here instead of `const` we could use cancelCase names but I just left it as is.

PR #1178, opened 2019-06-11
Extending BotModelMismatch - Rename to BotMissingApi

SDESkowronski:
- Renamed the BotModelMismatch test suite to BotMissingApi
- Extended the suite by adding new test scenarios to it
- Renamed some functions which names started with "validate" to be consistent and start with "verify"
- Clarified the names of a few other functions
- Updated TestGrid to show new test coverage


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -1338,7 +1338,7 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                                                         })}
                                                 </div>}
                                         </div>
-                                        : <div className="cl-errorpanel">
+                                        : <div className="cl-errorpanel" data-testid="action-creator-editor-error">
```
mattmazzola: I would think it's better to make this `data-testid` more specific to callback errors.  `data-testid="action-creator-editor-error"` could be any error, but this file has template errors and callback errors and I tests would want to distinguish.

SDESkowronski: I used that name since it matches the file name it resides in. That is the pattern I have established from the start. From my perspective it helps when trying to find what the id is associated with. However, I am not attached to that id name. So if you have a better name tell me what it is and I'll make the change.

mattmazzola: Yes, he name of the file is good candidate for the prefix of the test id, but the id has be specifc to what it's identifying otherwise it risks overlap and is ambiguous.

For this case (callback errors) I would suggest:
`data-testid="action-creator-editor-error-callback"`

And in the other template error you can use:
`data-testid="action-creator-editor-error-template"`

PR #1184, opened 2019-06-12
Api Memory Manipulation Test Suite

SDESkowronski:
Added a new test suite that uses APIs that manipulates Entity memory contents to verify the coupling of these various pieces all work together end-to-end. The basic pieces came from our Pizza Demo and the test scenario follows a simplified version of those demos.


SDESkowronski:
Merging into Master now as I got no response and Matt is out of office till Monday and I have another PR to submit. Will entertain any late comments and fix later if need be.

PR #1187, opened 2019-06-18
fix: rely on location.hostname instead of hardcoded localhost

msft-shahins:
This change enables serving the UI from a docker container or deployed application. It relies on `location.hostname` as the base url for SDK and also directline server. Also, it set the default port to port '80' if `location.port` is empty string.


src/types/const.ts
```diff
@@ -22,8 +22,12 @@ export enum SelectionType {
 
 export const CL_IMPORT_ID = '9c110735ea8b440d8f31c5c68ffb767d'
 
+export const LOCATION_HOST : Readonly<string> = location.hostname
+
+const LOCATION_PORT : string = location.port === '' ? '80' : location.port
```
mattmazzola: Can you explain the `LOCATION_PORT` change and in what case it was needed?

I had only considered these cases:

local dev - 3000 (or custom port by over ride)
local bot - 3978 (or custom port by over ride)

On, deployed bot's we don't run UI as it's a risk, but you could enable it. on. 

I'm not exactly sure, but on Azure internally it's a string such as "pipe://1234abdf23", and I assume other cloud providers have others, but this gets mapped out to 80 on the external facing URL.

I saw this was about using docker. Does the bot run on 80 in some cases?

src/types/const.ts
```diff
@@ -22,8 +22,12 @@ export enum SelectionType {
 
 export const CL_IMPORT_ID = '9c110735ea8b440d8f31c5c68ffb767d'
 
+export const LOCATION_HOST : Readonly<string> = location.hostname
```
mattmazzola: Hmm, is this planned to change in future and prevent the changes in dependent files?

Looks like just a constant mapped to other constant, was wondering if we could just put `location.hostname` in place of `LOCATION_HOSTNAME`.  My first thought if I saw `LOCATION_HOSTNAME` is "Wonder why this is here? Why aren't they using the browser's location object, must be something special" Then I see it is using `location.hostname` just hidden.

msft-shahins: The idea was making this configurable later. If UI wants to be served by a server that is different than the host that serves the bot, i.e. SDK and directline offline. probably better name is BOT_HOSTNAME instead of LOCATION_HOSTNAME. What do you think?

mattmazzola: Oh ok. Yea that name, bot hostname, is better.  

I'm thinking this could be another global setting similar to the existing bot port.
My understanding is this would allow running UI locally to target a bot that is deployed or really any difference as you said because it decouples the two further.

PR #1188, opened 2019-06-18
Create Entity Tests

SDESkowronski:
- Split the CreateModels/AllEntityTypes into two tests
- Got rid of the FlipFlop code - expanded to cover all combinations
- Expanded the suite to cover creating Pretrained Entities followed by Custom with same Resolver type as well as doing it in the reverse order to ensure that the informational popup only comes up once per Resolver type.


cypress/integration/Regression/CreateModels/AllEntityTypes2.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    })
+  })
+
+  // These tests differ from the tests in AllEntityTypes1 by reversing the order of creating the Entities
+  // with resolver type and pretrained, along with adding an extra resolver type entity. The purpose is to
+  // verify that we don't get the pop-up (after the first of that type) that says...
+  //    Note: You must wait for training to complete before the new pre-trained Entity will be detected
+  // Also we want to verify all combinations of multivalue and negatable resolver types can be created.
+  context('Create Entities', () => {
+    entities.pretrainedEntityTypes.forEach(entityType => {
```
mattmazzola: I didn't know if you planned to add them in separate PR which is fine, but we recently added `geographyV2`, `personName`, and `keyPhrase` as new types.

You would need to update the `pretrainedEntityTypes` list if you want to test those although it doesn't matter much as testing path of one is identical to another except the name.

cypress/integration/Regression/CreateModels/ApiCallbacks.spec.js
```diff
@@ -20,11 +20,11 @@ describe.skip('API Callbacks - CreateModels', () => {
 
   context('Create Entities', () => {
     it('Should create custom trained entities that will be used by the API Callbacks', () => {
-      entities.CreateNewEntity({ type: 'Custom Trained', name: '1stArg'})
-      entities.CreateNewEntity({ type: 'Custom Trained', name: '2ndArg'})
-      entities.CreateNewEntity({ type: 'Custom Trained', name: 'entityError', negatable: true})
-      entities.CreateNewEntity({ type: 'Custom Trained', name: 'logicError', negatable: true})
-      entities.CreateNewEntity({ type: 'Custom Trained', name: 'renderError', negatable: true})
+      entities.CreateNewEntityThenVerifyInGrid({ type: 'Custom Trained', name: '1stArg'})
```
mattmazzola: I would mention that expanding the function names like this is a sign of the function doing too many things. Ideally want atomic / independent things that compose together.  Example, you would also want to be able to verify grid even after an Update instead of Create, or be to create an entity within an Action and then go to entity page and verify it's in the grid correctly which injects navigation steps.  All these variations are not allowed when coded like this.

If you're going to keep doing this at least make the functions internally separate.
You have `CreateNewEntityThenVerifyInGrid` which calls `CreateNewEntity` but still has the verifying grid logic directly inside it on the same level.  Not important for this PR but in future I'd rather see:

- CreateNewEntityThenVerifyInGrid
  - CreateNewEntity
  - VerifyEntityInGrid

SDESkowronski: Okay, I'll do that soon.

PR #1191, opened 2019-06-19
 feat: remove training status from dialogs

mattmazzola:
Removes the training status component from the dialog modals
Doesn't fetch training status during scoring since it would not have changed.


src/components/modals/EditDialogAdmin.tsx
```diff
@@ -398,9 +397,7 @@ class EditDialogAdmin extends React.Component<Props, ComponentState> {
                         onAddTag={this.props.onAddTag}
                         onRemoveTag={this.props.onRemoveTag}
                     />
-                    <TrainingStatusContainer
-                        app={this.props.app}
-                    />
+                    <b></b>
```
mattmazzola: Note: Use of `<b>` tag here is convention I've seen to indicate it's just a placeholder for the grid.  Similar to `div` it has no effect on the page.

PR #1194, opened 2019-06-21
feat: delete selected log dialogs

mattmazzola:
Video:
https://microsoft.sharepoint.com/:v:/t/convsysresearch/EbWKgwVBXg9Iv22ot8fvpdUBc6ct1zRgUEMj96-rPeHh9Q?e=fYwxBd

Turns out there hidden `data-selection-invoke={true}` attribute you can apply to your `DetailsList` rows.

There was significant change to how we render the list due to needing the selection/checkboxes which makes the more complex.  

Before:
We would run all the log dialogs through a filter and sort function based on current filters and sort columns. I'm not sure why it mattered, but this kept reseting the selection on every render likely because it would be a new object of items (even though they were the same items).

After:
Now we have to update the list of items on every change that could impact them (search value, and sorting) which is where the complexity arises.


src/actions/logActions.ts
```diff
@@ -51,6 +51,44 @@ export const deleteLogDialogThunkAsync = (app: AppBase, logDialogId: string, pac
...
+    }
+}
+
+export const deleteLogDialogsThunkAsync = (app: AppBase, logDialogIds: string[], packageId: string) => {
+    return async (dispatch: Dispatch<any>) => {
+        dispatch(deleteLogDialogsAsync(app.appId))
+        const clClient = ClientFactory.getInstance(AT.DELETE_LOG_DIALOGS_ASYNC)
+
+        try {
+            await clClient.logDialogsDeleteMany(app.appId, logDialogIds)
```
LarsLiden: Since we don't block the UI, is there a point where user can open one of the log dialogs before the deleted is completed and end up in a weird state?

mattmazzola: Confirm this was possible.  I did what we talked about in person, moving the removal of the dialogs from the store so the `ASYNC` call instead of `FULFILLED` call so they happen preemptively.

PR #1198, opened 2019-06-24
Entity test scenarios

SDESkowronski:
- Changed "expectedEntities" (plural) to "expectedEntity" (singular) as that is the reality of the UI.
- Created new test suite "EntitiesActions/EditAndDeleteEntities"
- Fixed a bug I found in the "DoesNotContain" method and added a "DoesNotContainExact" version as well.


cypress/support/Entities.js
```diff
@@ -6,6 +6,21 @@
...
+  "number",
+  "ordinal",
+  "percentage",
+  "temperature",
+  "dimension",
+  "money",
+  "age",
+  "url",
+  "email",
+  "phonenumber",
```
mattmazzola: Maybe you intended not to add them, but since you're editing this anyways I would remind of the new types `geographyV2, personName, and keyPhrase`.

SDESkowronski: It was intentional because I wasn't ready to test this change, but I just now made and tested the change since it really is simple. I'll push that in with the rest of this code.

PR #1199, opened 2019-06-25
fix: prevent Cannot read property 'focus' of null

mattmazzola:
I'm seeing these errors in some of the tests. 
```
Uncaught TypeError: Cannot read property 'focus' of null

This error originated from your application code, not from Cypress.

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the 'uncaught:exception' event.

https://on.cypress.io/uncaught-exception-from-application
```

I'm guessing it is another race condition where most of the time the button is assigned before we call `focus()` but in some cases it isn't assigned yet.

I need to find a way to reproduce this and debug, but my guess is there is problem with this pattern:
```
// Wrong type - should be OF.IButton | undefined since it's temporarily in that state
newEntityButton: OF.IButton 
...
componentDidMount() {
     // call focus() but the button is still null or undefine
    this.newEntityButton.focus()
}
```

I meant to catch all of these when I changed it to use React.RefObjects but I missed these.

### DIFF
--- a/src/routes/Apps/App/Entities.tsx
+++ b/src/routes/Apps/App/Entities.tsx
@@ -145,7 +145,7 @@ interface ComponentState {
 }
 
 class Entities extends React.Component<Props, ComponentState> {
-    newEntityButton: OF.IButton
+    newEntityButtonRef = React.createRef<OF.IButton>()
     state: ComponentState
 
     constructor(props: Props) {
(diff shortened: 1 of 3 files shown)


mattmazzola:
Never mind, I think it is @conversationlearner/webchat code.

image

image

mattmazzola:
It seems Lars updated webchat to fix the issue.

PR #1201, opened 2019-06-25
fix: file import selector

mattmazzola:
I'm not quite sure what happened here.  I assume this broke because of change to 6.x office fabric where they removed a `div` yet it was said tests had been passing. You can see DOM doesn't have the div.  This selector was overly specific anyways.

`CypressError: Timed out retrying: Expected to find element: '[data-testid="model-creator-import-file-picker"] > div > input[type="file"]', but never found it.`

image


SDESkowronski:
Sorry for the late response. I fixed this yesterday and merged it into master.

PR #1223, opened 2019-07-08
fix: bug 2193 ignore card button click after end session

LarsLiden:


LarsLiden:
DialogMode is part of a teach session.  When editing there is no teach session just the history

PR #1237, opened 2019-07-19
feat: testing panel

LarsLiden:


src/components/modals/RateDialogsModal.tsx
```diff
@@ -0,0 +1,385 @@
...
+    render() {
+        const leftHistory = this.state.isFlipped ? this.state.history2 : this.state.history1
+        const rightHistory = this.state.isFlipped ? this.state.history1 : this.state.history2
+
+        return (
+            <div>
+                <OF.Modal
+                    isOpen={true}
+                    isBlocking={true}
+                    containerClassName="cl-modal cl-modal--compare-dialogs cl-modal--teach"
```
mattmazzola: This is not a teach modal. I would make another class that copies if you need those attributes. `cl-modal--rating` or something.  Would be cool to use this for merges.

src/routes/Apps/App/Testing.css
```diff
@@ -0,0 +1,79 @@
+.cl-testing-header {
```
mattmazzola: I find it hard to believe that these are all top level classes. I would assume most of them are children, would change them to follow the rest of the BEM classes. http://getbem.com/naming/

package.json
```diff
@@ -32,14 +32,16 @@
...
     "immer": "3.1.3",
     "immutable": "^3.8.2",
     "inline-css": "^2.4.1",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.15",
     "markdown-it": "^8.4.2",
     "moment": "^2.22.2",
     "node-zip": "^1.1.1",
     "office-ui-fabric-react": "6.185.0",
+    "query-string": "^6.8.1",
```
mattmazzola: Hmm, I also needed to do work with querystrings what did you need it for?

LarsLiden: My bad. Was originally going to use query string for flighting but moved to setting page.  Want me to leave it in for you?

package.json
```diff
@@ -90,7 +93,7 @@
     "@types/uuid": "^3.4.4",
     "commitizen": "3.1.1",
     "cross-env": "^5.2.0",
-    "cypress": "^3.1.5",
+    "cypress": "^3.4.0",
```
mattmazzola: Surprized you changed this.  I would think it's risky because it could introduce change which break tests but we should try to keep up to do date and would probably do it anyways.

package.json
mattmazzola: To late now it's already merged, might as well leave it

package.json
LarsLiden: That was from an audit fix.  Removed about 500+ high error warnings

mattmazzola: Yea, I'd prefer those be in a audit fix PR so it's clear, otherwise it looks like this package change was related to your testing panel

LarsLiden:
👍

package.json
LarsLiden: 👍

src/components/modals/RateDialogsModal.tsx
LarsLiden: fixed

PR #1242, opened 2019-07-22
style: edit modal uses query param

LarsLiden:
Not merging yet, but want your feedback.  Thanks!

### DIFF
--- a/src/routes/Apps/App/TrainDialogs.tsx
+++ b/src/routes/Apps/App/TrainDialogs.tsx
@@ -17,6 +17,8 @@ import actions from '../../../actions'
 import TreeView from '../../../components/modals/TreeView/TreeView'
 import TranscriptImporter from '../../../components/modals/TranscriptImporter'
 import TranscriptImportWaitModal from '../../../components/modals/TranscriptImportWaitModal'
+import { withRouter } from 'react-router-dom'
+import { RouteComponentProps } from 'react-router'
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import { returntypeof } from 'react-redux-typescript'
@@ -142,7 +144,7 @@ const defaultTagFilter = (intl: InjectedIntl) => ({ key: -1, text: Util.formatMe
 interface ComponentState {
     columns: IRenderableColumn[]
     sortColumn: IRenderableColumn
-    history: Activity[]
+    activityHistory: Activity[]
     lastAction: CLM.ActionBase | null
     isTeachDialogModalOpen: boolean
     isEditDialogModalOpen: boolean
@@ -192,7 +194,7 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         this.state = {
             columns: columns,
             sortColumn: lastModifiedColumn,
-            history: [],
+            activityHistory: [],
             lastAction: null,
             isTeachDialogModalOpen: false,
             isEditDialogModalOpen: false,
@@ -254,6 +256,33 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         // If train dialogs have been updated, update selected trainDialog too
         if (this.props.trainDialogs !== newProps.trainDialogs) {
             this.focusNewTeachSessionButton();
+        }
+    }
+
+    async componentDidUpdate(prevProps: Props, prevState: ComponentState) {
+        this.handleQueryParameter(prevProps, prevState)
+    }
+
+    async handleQueryParameter(prevProps: Props, prevState: ComponentState): Promise<void> {
+        const searchParamsPrev = new URLSearchParams(prevProps.location.search)
+        const selectedDialogIdPrev = searchParamsPrev.get("id")
+
+        const searchParams = new URLSearchParams(this.props.location.search)
+        const selectedDialogId = searchParams.get("id")
+
+        if (selectedDialogId === selectedDialogIdPrev) {
+            return
+        }
+
+        if (selectedDialogId && 
+            (!this.state.isEditDialogModalOpen && !this.state.isTeachDialogModalOpen)) {
+            const trainDialog = this.props.trainDialogs.find(td => td.trainDialogId === selectedDialogId)
+            if (!trainDialog) {
+                // Invalid train dialog, go back to TD list
+                this.props.history.goBack()
+                return
+            }
+            this.selectTrainDialog(trainDialog)
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -257,6 +259,33 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         }
     }
 
+    async componentDidUpdate(prevProps: Props, prevState: ComponentState) {
+        this.handleQueryParameter(prevProps, prevState)
+    }
+
+    async handleQueryParameter(prevProps: Props, prevState: ComponentState): Promise<void> {
+        const searchParamsPrev = new URLSearchParams(prevProps.location.search)
+        const selectedDialogIdPrev = searchParamsPrev.get("id")
```
mattmazzola: Would suggest to move "id" to a constant some where so it can't get out of sync with distance calls and it's more strongly typed since `URLSearchParams` isn't. I like that it uses the native browser `URLSearchParams` instead of using library though.

Maybe:
```
const queryParams = {
  id: "id"
}
```

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -142,7 +144,7 @@ const defaultTagFilter = (intl: InjectedIntl) => ({ key: -1, text: Util.formatMe
 interface ComponentState {
     columns: IRenderableColumn[]
     sortColumn: IRenderableColumn
-    history: Activity[]
+    activityHistory: Activity[]
```
mattmazzola: Glad to see this renamed. Never liked the original name as could be history of anything and browsers already have history.

What keeps it from being `activities`?

It would match what we do for other arrays like `actions`, `trainDialogs`, etc. I noticed the `history` variable in WebChat also and wondered if it was re-used from their code.

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -922,13 +951,18 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
...
         const selectedActivityIndex = DialogUtils.activityIndexFromRound(trainDialog, roundIndex, scoreIndex) || null
-        await this.onClickTrainDialogItem(trainDialog, EditDialogType.TRAIN_ORIGINAL, selectedActivityIndex)
+        await this.selectTrainDialog(trainDialog, EditDialogType.TRAIN_ORIGINAL, selectedActivityIndex)
     }
 
     @OF.autobind
-    async onClickTrainDialogItem(trainDialog: CLM.TrainDialog, editType: EditDialogType = EditDialogType.TRAIN_ORIGINAL, selectedActivityIndex: number | null = null) {
+    async onClickTrainDialogItem(trainDialog: CLM.TrainDialog) {
+        const { history } = this.props
+        let url = `/home/${this.props.app.appId}/trainDialogs?id=${trainDialog.trainDialogId}`
```
mattmazzola: Should be a way to use `match.url` here. Which is the current state of router. This way we can change parent paths such as home or trainDialogs an this will still work instead of being hard coded.

mattmazzola: Again, would use the constant mentioned above when generating the querystring so it can't get mismatched. Optional since it's simple case but generating the query string using one of the many libraries. We already depend on 3 different packages (queryString, querystring, query-string) which do the same thing so might as well use them.

Putting it all together I was envisioning something like this:
```
import * as queryString from 'query-string'
...
const queryObject = {
  [queryParams.id]: trainDialog.trainDialogId
}

const query = queryString.stringify(queryObject)

${match.url}?${query}
```

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -922,13 +951,18 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
...
+        history.push(url, { app: this.props.app })
+    }
+
+    async selectTrainDialog(trainDialog: CLM.TrainDialog, editType: EditDialogType = EditDialogType.TRAIN_ORIGINAL, selectedActivityIndex: number | null = null) {
```
mattmazzola: Don't feel too strongly but was wondering if there is better name `selectTrainDialog` ?

From what I can tell there are cases where we open the model even if the user didn't click on an item so onClick is no longer accurate and it makes sense to split these functions but they still seem to overlap. Select is very much like Click

Is the effect of this function showing a dialog in the modal?  Was thinking `viewTrainDialog` but then it sounds render related.

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -257,6 +259,33 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
...
+        if (selectedDialogId === selectedDialogIdPrev) {
+            return
+        }
+
+        if (selectedDialogId && 
+            (!this.state.isEditDialogModalOpen && !this.state.isTeachDialogModalOpen)) {
+            const trainDialog = this.props.trainDialogs.find(td => td.trainDialogId === selectedDialogId)
+            if (!trainDialog) {
+                // Invalid train dialog, go back to TD list
+                this.props.history.replace(`/home/${this.props.app.appId}/trainDialogs`, {app: this.props.app})
```
mattmazzola: Similar, should be able to use the `match.url` or something on the match object.

mattmazzola: I think it is `match.path` 
https://reacttraining.com/react-router/web/api/match

mattmazzola:
I should have noted, this is not a `style: ` type PR. The ability to refresh and preserve the item selected in a list a feature to me.

Would do something like `feat: preserve selected item on refresh`

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: Thought about that, although there is some value in the name "history" is that it represent the activities that existed before the new teach session started.  There could be new activities in the webchat

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: good suggestion

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: openLogDialog?

mattmazzola: Ok, I think that more clearly describes effect of function.  Also, not sure if it has effect, but you can make it `private` to further tell that the component should be using `onClickTrainDialogItem` in the JSX but `selectTrainDialog` internally.

PR #1243, opened 2019-07-23
feat: open log dialog from tester

LarsLiden:


src/components/modals/CompareDialogsModal.tsx
```diff
@@ -157,6 +160,14 @@ class CompareDialogsModal extends React.Component<Props, ComponentState> {
         })
     }
 
+    @OF.autobind
+    onEdit() {
+        const validationResult = this.props.transcriptValidationResults[this.state.resultIndex]
+        const { history } = this.props
+        let url = `/home/${this.props.app.appId}/logDialogs?${DialogUtils.DialogQueryParams.id}=${validationResult.logDialogId}`
```
mattmazzola: What was the opinion on the other PR of generating these query strings from the object?

I think it's more scalable and less error prone because it eliminates typos, is easier to add a prop, etc. I had intended to clean up the whole client doing that.

src/routes/Apps/App/LogDialogs.tsx
```diff
@@ -332,28 +333,40 @@ class LogDialogs extends React.Component<Props, ComponentState> {
...
             })
         }
-        this.handleQueryParameter(prevProps, prevState)
+        this.handleQueryParameters(this.props.location.search, prevProps.location.search)
     }
 
-    async handleQueryParameter(prevProps: Props, prevState: ComponentState): Promise<void> {
-        const searchParamsPrev = new URLSearchParams(prevProps.location.search)
-        const selectedDialogIdPrev = searchParamsPrev.get(DialogUtils.DialogQueryParams.id)
+    async handleQueryParameters(newSearch: string, oldSearch?: string): Promise<void> {
```
mattmazzola: Nice to see this changed

PR #1244, opened 2019-07-23
feat: add replay all button

mattmazzola:
Makes a PUT /trainDialogs/ request for reach dialog which is not great as it puts extra stress on server, but it does solve an annoyance with our product that the only other way is achieve this is to manually go through each dialog and save.

If we're worried we could extend with these options:
1. Determine if any dialogs have unsaved changes and only allow replay all if this is the case.
2. Add batching API so it can send multiple the dialogs in one call
3. Split out saving so users have to click save all after replay all and it's not automatically done.
4. Create confirmation pop-up window warning of saving more than N number of dialogs.


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -971,6 +973,29 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
         history.push(url, { app: this.props.app })
     }
 
+    @OF.autobind
+    async onClickReplayAll() {
+        this.setState({
+            isReplayAllInProgress: true
+        })
+
+        for (const trainDialog of this.props.trainDialogs) {
```
LarsLiden: Seems like it should just replay the train dialogs that need replay rather than all of them.

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -1046,13 +1072,13 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
...
 
-        this.setState({isImportWaitModalOpen: true})
+        this.setState({ isImportWaitModalOpen: true })
 
         const transcriptHash = Util.hashText(JSON.stringify(transcript))
 
         // If transcript has already been imported, go to next one
         if (this.hasTranscriptBeenImported(transcriptHash)) {
-            this.setState({isImportWaitModalOpen: false})
+            this.setState({ isImportWaitModalOpen: false })
```
LarsLiden: By the way, what do you have enabled to autofix spacing issues?

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -1338,6 +1367,15 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
                             text={Util.formatMessageId(intl, FM.TRAINDIALOGS_TREEVIEW_BUTTON)}
                         />
                     }
+                    <OF.DefaultButton
```
LarsLiden: I'd disable or hide this button when it's not needed?   It's kinda confusing.   Maybe a warning when you click on it?  User is tarting a big operation and may have no idea what they are about to do

src/routes/Apps/App/TrainDialogs.tsx
mattmazzola: Yes, what is reliable way to determine that state?  I assumed checking for invalid state was not enough, because I thought it was possible that any dialog that needs replay for other reason that being invalid.

There was similar proposal in the description:

> Determine if any dialogs have unsaved changes and only allow replay all if this is the case

If we can determine if dialogs need replay we can also do this.

src/routes/Apps/App/TrainDialogs.tsx
mattmazzola: image

I use `Ctl + Alt + F`, I looked at enabling `prettier` on commit before, but it wasn't produce desirable formatting, although from this PR it looks like I'll have to look at that again becuase this seems be happening more and more where there are many distracting changes not related to the PR.

LarsLiden:
I think UNKNOWN and WARNING are the only states where replay is required, though if user made an API change they want to test…  Hmm

src/routes/Apps/App/TrainDialogs.tsx
mattmazzola: Yes, both of these were in the description as well.  We need to be able to determine when it's not needed so it's related to that question.

And yes, what do you consider a big operation 2 or more dialogs? or any replay all? or 10 dialogs?

src/routes/Apps/App/TrainDialogs.tsx
mattmazzola: I should add, that if we do this change which we would want to do for performance we can't label the button "Replay All" because it's not guaranteed to be "All" dialogs anymore.  I'm not sure of a good name. We can't say "Replay Invalid" either.

Perhaps we have to do the work for selection and add "Replay Selected"

LarsLiden:
Yeah – good call.  I like the replay selected.   Give more control

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: If we switch to "replay selected" I don't think disable makes sense.   An alternative to the warning would be a progress modal with a cancel button that lets user quit in the middle.  I'm using that for the testing stuff

LarsLiden: See [TestWaitModal.tsx](https://github.com/microsoft/ConversationLearner-UI/blob/master/src/components/modals/TestWaitModal.tsx)

mattmazzola:
Ok, replay selected:

src/routes/Apps/App/TrainDialogs.tsx
LarsLiden: Any thoughts on the above?  Think that would help with folks not knowing they may be starting something that could take 20 minutes?   Something like this but "Replaying" instead of "Testing"

image

mattmazzola: I didn't realize this could take 20 minutes, only thought it was a lot of network calls. I suppose we don't know how long their Code callbacks will take and they could take very long time.

Anyways, will try to add this.

LarsLiden: Yeah, testing can take a while at it has to replay every test dialog through very step (entity extraction, scorer step, etc) and you may be testing 100+ .transcript files

mattmazzola: Hmm, looks like the larger spinner blocks interaction with the modal anyways. I'm guessing since the testing functions were new they didn't have dependent logic to work around.

I see two options:
1. remove the modal and go with as 
2. add some other flag to the action to not enable the spinner during this special type of replay
Seems like we keep adding these flags for special conditions which adds a lot of complexity. Seems like we should remove the global at some point if we're going to combine it with these local modal spinner.

mattmazzola: Well went ahead with #2.

PR #1246, opened 2019-07-24
fix: null ref exception in actioncreatoreditor

LarsLiden:


mattmazzola:
There's no description, why is this value becoming `null`? That seems to be the issue we need to solve, not just preventing null from being assigned to an array.

PR #1256, opened 2019-08-08
feat: use DISPATCH actions instead of overloading TEXT actions

mattmazzola:
Depends on the SDK understanding these new types, and the server and trainer accepting them.

### DIFF
--- a/src/routes/Apps/AppsIndex.tsx
+++ b/src/routes/Apps/AppsIndex.tsx
@@ -146,8 +146,7 @@ function generateDispatcherSource(sourceModels: SourceModel[]): AppDefinition {
         "negativeConditions": [],
         "clientData": {
             "importHashes": []
-        },
-        "modelId": sourceModel.model.appId,
+        }
     }))
 
     /**
--- a/src/routes/Apps/AppsIndex.tsx
+++ b/src/routes/Apps/AppsIndex.tsx
@@ -136,8 +136,8 @@ function generateDispatcherSource(sourceModels: SourceModel[]): AppDefinition {
     const actions = sourceModels.map<any>(sourceModel => ({
         "actionId": uuid(),
         "createdDateTime": new Date().toJSON(),
-        "actionType": "TEXT",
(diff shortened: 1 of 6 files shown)


src/routes/Apps/AppsIndex.tsx
```diff
@@ -136,8 +136,8 @@ function generateDispatcherSource(sourceModels: SourceModel[]): AppDefinition {
     const actions = sourceModels.map<any>(sourceModel => ({
         "actionId": uuid(),
         "createdDateTime": new Date().toJSON(),
-        "actionType": "TEXT",
-        "payload": `{\"json\":{\"kind\":\"value\",\"document\":{\"kind\":\"document\",\"data\":{},\"nodes\":[{\"kind\":\"block\",\"type\":\"line\",\"isVoid\":false,\"data\":{},\"nodes\":[{\"kind\":\"text\",\"leaves\":[{\"kind\":\"leaf\",\"text\":\"@DISPATCH:${sourceModel.model.appId}:${sourceModel.model.appName}\",\"marks\":[]}]}]}]}}}`,
+        "actionType": "DISPATCH",
+        "payload": `{\"modelId\": \"${sourceModel.model.appId}\", \"modelName\": \"${sourceModel.model.appName}\"}`,
```
mattmazzola: We store `modelName` here because we needed something useful to display to the user instead of a GUID, however we would ideally only store the `modelId` because the Model name can be changed and then out of date.

However, there are many places we use the modelName which are not asynchronous or have easy ability models such as inside ActionDetailsList, and synchronous locations.

LarsLiden: Hmm...  Won't this break (or at least show the wrong model name) if the user does a rename?  Is the issue that we need a lookup the model name?

mattmazzola: I'm not sure what the question is. It seems to restate what the above post said.

> we would ideally only store the modelId because the Model name can be changed and then out of date.

Yes, ideally we wouldn't store name, but the effort to go refactor all these locations to allow asynchronous rendering which would likely require a loading state until the request resolved and then accounting for failure, etc is not worth it if it turns out this whole concept if Dispatching actions doesn't work in current form.

PR #1258, opened 2019-08-14
Import obi dialgos

LarsLiden:


src/Utils/dialogUtils.tsx
```diff
@@ -261,9 +261,13 @@ export function dialogSampleInput(dialog: CLM.TrainDialog | CLM.LogDialog): stri
     let round = 0
     let length = 0
     while (round < dialog.rounds.length && length < MAX_SAMPLE_INPUT_LENGTH) {
-        const userInput =
-            (dialog as CLM.LogDialog).rounds[round].extractorStep.text ||
-            (dialog as CLM.TrainDialog).rounds[round].extractorStep.textVariations[0].text
+        let userInput = (dialog as CLM.LogDialog).rounds[round].extractorStep.text
+        if (!userInput && (dialog as CLM.TrainDialog).rounds[round].extractorStep.textVariations[0]) {
+            userInput = (dialog as CLM.TrainDialog).rounds[round].extractorStep.textVariations[0].text
```
mattmazzola: Given `(dialog as CLM.TrainDialog).rounds[round].extractorStep` is repeated 3 times consider saving to variable.

The type assertion doesn't actually check if it's a log or train dialog round only forces it.

package.json
```diff
@@ -29,6 +29,7 @@
     "axios": "^0.19.0",
     "botbuilder": "4.1.7",
     "botframework-directlinejs": "^0.9.17",
+    "clone": "^2.1.2",
```
mattmazzola: What is this used for?

src/Utils/obiDialogParser.ts
```diff
@@ -0,0 +1,331 @@
...
+                                    // Add extractor step to all the children
+                                    childDialogs.forEach(td => {
+                                        td.rounds[0].extractorStep = extractorStep
+                                    })
+    
+                                    // Add children to train dialog list
+                                    trainDialogs = [...trainDialogs, ...childDialogs]
+                                }
+                                else if (step.$type === "Microsoft.SendActivity") {
+
```
mattmazzola: Something supposed to happen here? This is weird to be empty. Would at least put a comment saying it's intentionally skipping this `$type`, will be added later, or if expected not to happen do some sort of `throw new Error('Not Implemented')`

src/Utils/obiDialogParser.ts
```diff
@@ -0,0 +1,331 @@
...
+                }
+            }
+        }
+        let trainRound: CLM.TrainRound | undefined
+        if (obiDialog.steps) {
+            for (const step of obiDialog.steps) {
+                if (typeof step === "string") {
+                    throw new Error("Unexected step of type string")
+                }
+                else if (step.$type === "Microsoft.SendActivity") {
```
mattmazzola: Would move these to some `ObiStepType` enum.

src/routes/Apps/AppsListComponent.tsx
```diff
@@ -313,6 +308,25 @@ export class Component extends React.Component<Props, ComponentState> {
             </div>
         </div>
     }
+
+    private getSortedApplications(sortColumn: ISortableRenderableColumn, apps: CLM.AppBase[]): CLM.AppBase[] {
```
mattmazzola: Why was this moved? The whole ordering of public, private stuff seems like a weaker convention to follow than the stronger convention of having `render` be the last method in `React`. Not everyone is using TypeScript or private. Moving code is harder for code review / diffs as you can't tell was it moved and changed for just moved.

src/types/obiTypes.ts
```diff
@@ -0,0 +1,571 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+ */
+///////////////////////////////////
+// GENERATED FILE from OBI schema
```
mattmazzola: What the instructions to regenerate this? Would put them in the comment here.

package.json
LarsLiden: Nice catch.  Had to install it for some BF stuff that I ended up removing.

src/routes/Apps/AppsListComponent.tsx
LarsLiden: If we make that choice we should add an exclusion rule is TSLINT so that it stops showing errors

src/Utils/obiDialogParser.ts
```diff
@@ -0,0 +1,331 @@
...
+        if (obiDialog.rules) {
+            for (const rule of obiDialog.rules) {
+                if (rule.$type === "Microsoft.IntentRule") {
+    
+                    const textVariations = this.getTextVariations(rule.intent!)
+                    const extractorStep: CLM.TrainExtractorStep = {
+                        textVariations: textVariations
+                    }
+    
+                    if (rule.steps) {
```
tpmsr: This function gets pretty deeply nested.  How about putting this at the top of the for loop as `if (!rule.steps) continue` ?

src/Utils/obiDialogParser.ts
```diff
@@ -0,0 +1,331 @@
...
+        }
+        return trainDialogs
+    }
+    
+    private async getScorerStepfromActivity(prompt: string): Promise<CLM.TrainScorerStep> {
+    
+        const parsedActivity = prompt.substring(prompt.indexOf("[") + 1, prompt.lastIndexOf("]")).trim()
+        let response = this.composerDialog.lgMap.get(parsedActivity)
+        if (!response) {
+            // LARS temp to handle badly create dialog
```
tpmsr: Suggest rephrasing as a concrete TODO

(eg, `// TODO : throw Error once parsing is implemented for all activity types`)

...and removing the commented-out throw below

src/Utils/obiDialogParser.ts
```diff
@@ -0,0 +1,331 @@
...
+                    payload: action.payload,
+                    isTerminal: action.isTerminal,
+                    actionType: CLM.ActionTypes.TEXT,
+                    score: 1
+                }
+            }
+        }
+    
+        let scoreInput: CLM.ScoreInput = {
+            filledEntities: [],  //LARS handle filled entities from api calls
```
tpmsr: Did you means to check the `//LARS` comments in ?

tpmsr:
Mostly nits and some questions.  Looks like I was too late to catch you before the merge.

src/Utils/obiDialogParser.ts
LarsLiden: Yeah, they are a way for me to find actionable TODOs for myself.  We have a bunch of aspirational TODOs sprinkled in the code that make it hard to find things

PR #1259, opened 2019-08-14
Score Action Test Refactoring

SDESkowronski:
- Primary change is code to verify the Score Actions Pane of the Train Dialogs Modal
- 4 existing tests modified to use the new helper class "scorerModal.GeneratedData"
- Code refactoring to support these changes
- Miscellaneous test code bug fixes


cypress/support/components/ActionModal.js
```diff
@@ -51,6 +51,13 @@ export function TypeApiLogicArgs(args) { TypeApiArgs('Logic Arguments', args) }
 export function TypeApiRenderArgs(args) { TypeApiArgs('Render Arguments', args) }
 
 function TypeApiArgs(apiArgLabel, args) {
+  function ClickOnNoOpHack() { 
+    // This has no effect on the data in this Action Modal but it some how
+    // resets something in the UI that enables picking an entity.
+    // Bug 2132: TEST BLOCKER: Automation cannot trigger 2nd Entity picker in API Action arguments
+    cy.get('[data-testid="action-creator-wait-checkbox"]').click()
```
mattmazzola: I expect this to click twice to preserve the state of checkbox.
Clicking only once would seemingly leave it changed which is not a No Op and does affect the action.

cypress/fixtures/scoreActions/disqualifyingEntities.json
```diff
@@ -0,0 +1,446 @@
+[
```
mattmazzola: These seem like it data for the tests. The fixtures are supposed to be fixed responses for network requests, or other operations like file loading which we use to fake the `.cl` import.

cypress/support/components/ScorerModal.js
```diff
@@ -3,76 +3,116 @@
...
+
+export class GeneratedData {
+  constructor(dataFileName) {
+    this.dataFileName = dataFileName
+    this.generateScoreActionsData = Cypress.env('GENERATE_SCORE_ACTIONS_DATA')
+
+    if (this.generateScoreActionsData) {
+      this.data = []
+    } else {
+      context('Load Generated Verification Data', () => {
```
mattmazzola: Would just import this file as any other javascript file.

`import myTestData from './myTestData'`

Then use it as normal. Having all the `cy.readFile` and `cy.fixture` just adds complexity.

cypress/integration/Regression/Train/DisqualifyingEntities.spec.js
```diff
@@ -169,5 +140,6 @@ describe('Disqualifying Entities - Train', () => {
     })
   })
 
+  generatedScoreActionsData.SaveGeneratedData()
```
mattmazzola: Why does this save data out?
This could seemingly create loop where test saves out incorrect data for the next test run. It should just be static files no reading or writing needed.

Even so, it seems the test data is loaded which represents expected values in actions scores list, then verifies that it matches.

If it matches, then there is no value in saving since it would only be over writing the same values.
If it's different, then we have the issue above where the file unknowingly changes as tests run which means they're different on different peoples machines

cypress/support/components/ScorerModal.js
```diff
@@ -3,76 +3,116 @@
...
+  // acceptableScoreDeviation as the percentage points that we can allow the score to deviate and still be acceptable.
+  VerifyScoreActionsList(acceptableScoreDeviation = 10) {
+    if (this.generateScoreActionsData) {
+      it('GENERATE the Score Actions data', () => {
+        cy.wait(2000)
+        cy.WaitForStableDOM().then(() => { this.data.push(GenerateScoreActionsDataFromGrid()) })
+      })
+    } else {
+      it('Verify the Score Actions data', () => {
+        cy.WaitForStableDOM().then(() => VerifyScoreActions(this.data[this.index++], acceptableScoreDeviation))
```
mattmazzola: Strongly advise against increment or decrement operators in unless `for` loops

But regardless any sort of index mutation in here is bad.  In other words making this an explicit `this.index = this.index + 1` wouldn't help.

From what I understand, this makes this function behave differently implicitly based on call order. This is very confusing and requires knowledge of previous calls. Your goal should be to make these independent and reduce the knowledge people need to know to understand expected behavior.

cypress/support/components/ActionModal.js
SDESkowronski: It turns out that it works because I actually have to click on a child element (I believe) to get it to change the value in the check box. I've tested it this way and it works.

cypress/fixtures/scoreActions/disqualifyingEntities.json
SDESkowronski: Yes this is data for the tests. The FIXTURES folder for it seems like an appropriate place to me. I created a sub folder to keep these separate from the .cl files.

According to the Cypress documentation at https://docs.cypress.io/api/commands/fixture.html fixtures are: "a fixed set of data located in a file."

cypress/support/components/ScorerModal.js
SDESkowronski: While I did not realize that would work, and am not sure it really would work, my implementation is different for a reason.

The tests that use these can be run in two ways;
1. Generate mode which generates these data files.
2. Test mode which is the main mode used to test the UI and it consumes these files.

In my opinion, trying to do it the way you suggest would be more complex.

cypress/integration/Regression/Train/DisqualifyingEntities.spec.js
SDESkowronski: This saves data out when the environment variable "CYPRESS_GENERATE_SCORE_ACTIONS_DATA" exists on the machine. It is a developer setting. It is used during the development phase of creating and maintaining the test suite.

cypress/support/components/ScorerModal.js
SDESkowronski: Again this is due to the dual mode that the test can opperate in, the development mode or the test mode. The data is generated in the same order it is consumed in, thus this system works.

Useage of this class is simple and straight forward when you read the file instead of viewing it as a series of changes and read the usage examples contained in working tests that use it.

HOWEVER, I am interested in learning why you advise against increment or decrement operators. I am not aware of any real pitfalls with them.

PR #1260, opened 2019-08-15
feat: enable ability to regenerate train dialogs for dispatch model

mattmazzola:
I had already done 80% of the work for regen and ended up just finishing it so it at least compiles and does regenerate.  Due to the number of network requests this likely wouldn't stay in current implementation, but there were a lot of other little things included that I thought were important if someone wanted to pick up where I left off, or merge and continue later.

### Clean up of app metadata
  Was previous making some weird encoding of dispatcher type to detect format, then if the format was known to extract model ids, but I realized we already have that information captured in all the actions.
Now it only, reconstructs source models form the model id in the DISPATCH actions. This is also more accurate way as it can't get out of sync (someone edits markdown)

### Adds more comments and cleans up algorithim
Given models A, B, C. I had intended if starting in model A, to only attempt transition to other models (B or C); however, due to the flattening I noticed the algorithm can attempt to transition to itself.  A - A.  I was going to change this, because it is extra noise, but thought it could perhaps be good as it re-enforces to not transition.

### Moves algorithm to `dispatchUtils`
Given that we planned to add many algorithms I planned to have this file dedicated for have each type. I hoped there could be common signature they comply two, but as long is takes in `SourceModelPairs[] => AppDefinition` it should be acceptable.


src/routes/Apps/AppsIndex.tsx
```diff
@@ -14,10 +12,11 @@ import { returntypeof } from 'react-redux-typescript'
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import { State } from '../../types'
-import { AppBase, AppDefinition, TrainDialog } from '@conversationlearner/models'
+import * as CLM from '@conversationlearner/models'
 import { CL_IMPORT_TUTORIALS_USER_ID } from '../../types/const'
 import { OBIImportData } from '../../Utils/obiUtils'
-import { SourceModel } from '../../types/models';
+import * as DispatchUtils from '../../Utils/dispatchUtils'
```
LarsLiden: Nit:  Would move these up by other * imports

PR #1262, opened 2019-08-16
fix: Update obiDialogParser to use code from CL-Model

tpmsr:
Small change to use CL-Model code in https://github.com/microsoft/ConversationLearner-Models/pull/198 (addToLGMap -> CLM.ObiUtils.parseLGString)


tpmsr:
Yes - I was waiting for an LGTM on the Model PR first  :)

package.json
```diff
@@ -21,7 +21,7 @@
   },
   "private": true,
   "dependencies": {
-    "@conversationlearner/models": "0.208.0",
+    "@conversationlearner/models": "^0.208.1",
```
LarsLiden: Shouldn't be checked in with "^"

PR #1265, opened 2019-08-21
feat: lg support for transcript imports

LarsLiden:


src/Utils/obiDialogParser.ts
```diff
@@ -202,23 +189,11 @@ export class ObiDialogParser {
         if (!response) {
             // LARS thow error once CCI .dialog transformer has been fixed
             response = { text: "Can't Parse LU", suggestions: []}
```
tpmsr: LG

src/Utils/obiTranscriptParser.ts
```diff
@@ -0,0 +1,78 @@
...
+                throw new Error(`Expecting .transcript file.\n\n Given: ${transcriptFile.name}`)
+            }
+
+            let source = await Util.readFileAsync(transcriptFile)
+            try {
+                const transcript: BB.Activity[] = JSON.parse(source)
+                const transcriptHash = Util.hashText(JSON.stringify(transcript))
+    
+                // If transcript has already been imported, skip it
+                if (!this.hasTranscriptBeenImported(transcriptHash)) {
```
tpmsr: I have a mild preference for

> if (this.hasTranscriptBeenImported(transcriptHash)) {
>   continue
> }

This better matches the comment above, and obviates the extra level of indent, which is arguably less error-prone.

src/Utils/obiUtils.tsx
```diff
@@ -83,12 +83,45 @@ export function isSameActivity(activity1: BB.Activity, activity2: BB.Activity):
...
+            }
+        }
+    }
+    return lgMap
+}
+
+export function substituteLG(text: string, lgMap: Map<string, CLM.LGItem> | null): string {
+    
+    // If not LG substitution just return the text
+    if (!lgMap || !text.startsWith('[') || !text.endsWith(']')) {
```
tpmsr: These should be regularly-formed, so something like this would be safer :

```
const p = RegExp(/^\[\w+\]$/)
if (!lgMap || !p.test(text)) {
  return text
}
```

src/Utils/obiUtils.tsx
```diff
@@ -83,12 +83,45 @@ export function isSameActivity(activity1: BB.Activity, activity2: BB.Activity):
...
+        return text
+    }
+    const lgName = text.substring(text.indexOf("[") + 1, text.lastIndexOf("]")).trim()
```
tpmsr: Adding a capture group to the pattern above makes this a 1-shot:

```
p = RegExp(/^\[(\w+)\]$/)
match = p.exec(test)
if (!lgMap || !match) {
  return text
}
const lgName = match[1]

PR #1271, opened 2019-08-27
fix: add lightDismiss to ErrorPanel

mattmazzola:
Fixes ADO#2247

Not sure if this was missed or done intentionally before. Suppose we have to be sure want this.


src/components/HelpPanel.tsx
```diff
@@ -21,15 +21,12 @@ class HelpPanel extends React.Component<Props> {
     render() {
         return (
             <OF.Panel
-                focusTrapZoneProps={{}}
```
LarsLiden: If memory serves me you added this (and the close button) to fix an issue where the panel could get stuck open and wasn't closable in some circumstances.  It may be that this issue has been fixed in the latest version of Fabric, but we should keep an eye on it

mattmazzola: Yea, I couldn't remember if the dismissal type was intentional or not, but seems like it wasn't.

PR #1272, opened 2019-08-27
feat: add strict resolver option

mattmazzola:
Fixes ADO#2175

Disabled when no resolver is selected
image

image
image

Help text (will probably change depending on future decisions):
image


mattmazzola:
Ok, updates after recent talks.

Since we enable the checkbox by default we also have to disable it when resolver is changed back so logic checked / enabled when not NONE, unchecked / disabled when NONE, can be toggled when resolver is selected.
image

Option is none editable.

Help Panel for "Resolution Required" option
image

PR #1274, opened 2019-08-29
 feat: enable dispatch algorithm selection

mattmazzola:
## Enable algorithm selection on Dispatcher Creator modal

image

- Move to dedicated modal since the other is very overloaded with unrelated concerns.
- Re-organized / re factored algorithm code (`dispatchUtil`) to be more readable and have better facades points.
- Renamed existing to DeterministicSingleTransfer
- Added RandomSingleTransfer
- Note: RandomMultiTransfer is not implemented. Need to go over it with Lars as it's very different from the first two. (Think it was select N rounds subset from random dialogs and concat to larger dialog. Will likely be hard to tell if it's written correctly)

Interesting idea during algo design that I realized the number of transition points within a dialog could be dependent on the dialog and instead of be fixed value like (Attempt up to 5 transition points).

E.g. A small dialog of 7 might try to transition in 4 places
but a dialog of 100 rounds would need to transition on probably 20 places.
 
I still think it's eventually going to need text analysis to do this well. In other words transitioning is much more likely and relevant at certain points in the conversation than others and lends to ML problem.

### DIFF
--- a/src/Utils/dispatchUtils.ts
+++ b/src/Utils/dispatchUtils.ts
@@ -2,14 +2,27 @@ import * as CLM from '@conversationlearner/models'
 import { SourceAndModelPair } from "src/types/models"
 import * as uuid from 'uuid/v4'
 import * as Util from './util'
+import { DispatcherAlgorithmType } from 'src/components/modals/DispatcherCreator';
 
 export function generateDispatcherSource(
+    sourceModelPairs: SourceAndModelPair[],
+    algorithmType: DispatcherAlgorithmType,
+) {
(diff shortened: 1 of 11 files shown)


mattmazzola:
Oh also forgot, during refactor I notice my original implementation actually attempt transition _within_ the same model.  In the revised design this is removed because the original dialogs are already forwarded to the dispatcher model and the intra transfers was confusing to me, but I thought it could potentially add value. It wuld be a Input -> Input combination that has never been seen before but yet might be possible.

PR #1278, opened 2019-09-03
fix: undefined enum value id

mattmazzola:
Need models PR, use ensure enum values have ID's


mattmazzola:
Don't know how these files keep getting unformatted, I thought I went through all of them at one point. Anyways Shahin didn't want change on server so we make two calls here 

Fixes ADO#2259
Fixes ADO#2243

PR #1284, opened 2019-09-05
fix: bad card match

LarsLiden:


LarsLiden:
Re-submit PR.  Appveyor trigger failed

PR #1287, opened 2019-09-05
fix: remove no actions return

mattmazzola:
I had messaged this on ADO but just creating PR for visibility.

What did you mean by
> TODO: Handle reload when actions not yet loaded

It seemed to work fine without this, and it prevents the bug. Does  this protect against another issue?

Also, need to consider same code in LogDialogs page


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -305,11 +305,6 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
             }
         }
 
-        // TODO: Handle reload when actions not yet loaded
```
LarsLiden: Don't remember.  My guess would be that it refers to page reload (i.e. browser) and prevents a null ref.  I'd be hesitant to take it out unless it is causing other issues.

mattmazzola: It is causing other issues which why I am asking about it.  I think the alternative is to prevent the server from creating the dialog.  I think there was mechanism to show train dialogs even though the user never saved it to preserve their work hoping they could finish it later.

There is mismatch in that when creating dialog manually we require an action to save, but server does not require an action, this then makes the dialog un openable.

LarsLiden:
Gotcha.  I’d say just test the browser reload and see if it causes issues.  (It’s an edge case)

PR #1289, opened 2019-09-09
feat: add random multi transfer dispatch algo and test data option

mattmazzola:
- Adds new "Random multi Transfer" and "Test" option to Dispatcher creation modal
- Currently the test option uses algorithm is the same as that option but is separate so it could be made independent in the future. It just generates a model so we can export the .transcript files from it.

### Possible Improvements
Currently understanding is there is no difference between few large N dialogs with M transfers and many small N dialogs with fewer M transers. (number of transfers is same for both cases) For this reason the algorithm generates the max 100 round dialogs. The defaults are 10 dialogs with 20 transfers within them. Which means 5 rounds per transfers. 
For our source models where the source dialogs only have 5 rounds that means each transfer includes the whole dialog. When the algorithm picks the random location where there are less than 5 rounds remaning it just repeats those until it has taken 5. I'm not sure if this is good or not, but important point.

- Rounds selected based on source dialog length, instead of fixed? Hard to get right
- Number of transfers based on number of source models. E.g. If there are only 2 source models there should be good coverage within 1 dialog, if there at 20 source models we have exponentially many more possible transfers and need more dialogs

image


src/Utils/dispatchUtils.ts
```diff
@@ -238,28 +269,90 @@ function randomSingleTransfer(
...
+            const dialogsInModel = modelsTrainDialogs[modelIndex]
+            const dialogIndex = Math.floor(Math.random() * dialogsInModel.length)
+            const dialog = dialogsInModel[dialogIndex]
+
+            // Get random N consecutive rounds from dialog
+            // Substract from N to ensure there is space for slice, otherwise use at least 1 round
+            const maxIndex = Math.max(1, dialog.rounds.length - numberOfRoundsPerTransition)
+            const startRoundIndex = Math.floor(Math.random() * maxIndex)
+            let newRounds = dialog.rounds.slice(startRoundIndex, numberOfRoundsPerTransition)
+            // TODO: Decide if there is better alternative. Dialog could not have as many rounds as requested but still be valid
```
mattmazzola: The part I was mentioning about repeating when there are less than desired number of rounds.
In other words dialog has 9 rounds. Algorithm randomly selects position 6 to start and take 5 more rounds. There are only 3 so it just repeats [6,7,8,6,7].

PR #1292, opened 2019-09-09
feat: add html serializer and rules for Slate

mattmazzola:
Adds another way to create Slate values from HTML instead of from JSON.

Example is in the comments:
```
<p>This required <b data-id="1f99bb7a-41af-46bb-b2f0-9c5c23404266" data-name="myEntity">$myEntity</b> and <i>[optional <b data-id="7d27f6a3-566a-487f-9345-79ea1db62708" data-name="otherEntity">$otherEntity</b> ]</i> ending.</p>
```

Creates this:
image

Could store this on the server if we wanted, although I'm not sure if it has much more value. Maybe slightly more readable but still should not have server modifying it.


mattmazzola:
Tried to add some tests, but I can't get it to work and spent too much time not getting anywhere. It works when I test it in the browsers as thats how I was able to create the picture in PR description but when I tried in the test it fails to deserialize and thinks it's an empty input.

mattmazzola:
Ok, will work on adding that.

PR #1301, opened 2019-09-16
fix: use same case on feature string matching

mattmazzola:
There was utility added that used `toUpperCase()` on one side but not the other.
Purpose is to eliminate casing from matching for both the features and feature string. User shouldn't have to remember if it's "cci" "Cci", or "CCI", they all should work and easier to explain to people.


src/Utils/util.ts
```diff
@@ -271,7 +271,7 @@ export function hashText(text: string) {
 }
 
 export function isFeatureEnabled(featureString: string | undefined, feature: Const.FeatureStrings) {
-    if (featureString && featureString.toUpperCase().includes(feature)) {
+    if (featureString && featureString.toUpperCase().includes(feature.toUpperCase())) {
```
LarsLiden: Was assuming the enum definitions would all be defined as upper case (but can't hurt)

PR #1302, opened 2019-09-17
 feat: highlight entities and values in action lists

mattmazzola:
Fixes ADO#1651

- Added new payload renderer components to mimic the existing ones, but these render with highlights
- Changes ActionDetailsList and ActionScorer to use these new Components
- Remove " (quotation) marks wrapping Action arguments (payloads in Code or Card actions)

I kept the old components in there although I don't think they are used anymore and we could remove them if people are Ok with it. I generally prefer removing all non-used code but seems like there might be use case in future that needs to display an Action as text instead of with highlights and would be easier if it was still there.

New rules

type | status | color
---- | ------ | -----
Required | unfilled | Red
Required | filled | Green
Optional | unfilled | Blue
Optional | filled | Green

Example:
image

Note: Tried out using the `classnames` package for styling and might apply pattern in other places where it could help. https://reactjs.org/docs/faq-styling.html#how-do-i-add-css-classes-to-components


src/components/ActionDetailsList.tsx
```diff
@@ -228,38 +228,38 @@ export default connect<typeof stateProps, {}, ReceivedProps>(mapStateToProps, ma
 function getActionPayloadRenderer(action: CLM.ActionBase, component: ActionDetailsList, isValidationError: boolean) {
     if (action.actionType === CLM.ActionTypes.TEXT) {
         const textAction = new CLM.TextAction(action)
-        return (<ActionPayloadRenderers.TextPayloadRendererContainer
```
mattmazzola: This pattern follows for most of the components. Switch from `Container` to `WithHighlights` and add new property to control marking of missing entities. (ActionDetailList no memory so there is no concept of missing, but in TrainDialog where memory exists we do need to know this)

src/components/modals/ActionPayloadEditor/slateTransformer.ts
```diff
@@ -46,14 +47,108 @@ function updateNodeOptionNames(node: any, options: IOption[], newName?: string)
...
+ * Replaces entity nodes with values. The values name be entity names, memory values, etc.
+ * This is more flexible than the other functions which only replace with fixed thing name or memory and cant recognize missing entities
+ * Hopefully the other functions will be retired and replaced in future.
+ *
+ * @param node Slate node (Value, Document, Block, Inline, Text, etc)
+ * @param entityValuesMap Map from entityId to entityEntry which has name and ?value
+ * @param replacerFn Function to be run on each replacement of entity node. Can be anything, but usually replaces with entity name or memory value
+ * @param entitiesRequired Boolean indicating if `entityMissing` flag should be added when entity without memory value is present
+ */
+function replaceEntityNodesWithValues(node: any, entityValuesMap: Record<string, Util.EntityMapEntry>, replacerFn: (entityEntry: Util.EntityMapEntry) => string, entitiesRequired: boolean): any {
```
mattmazzola: I think it wold be more appropriate to do this operation in `Slate` domain instead of JSON since they have their own API and normalizers than ensure allowed operations are done, but I ended up doing in JSON domain.

Might need to revisit this in future.

mattmazzola:
> Is it possible to add unit tests for any of this ?

Yes, we have `jest` setup and there are plenty of libraries to help with this. When you say "any" we have unit tests for some of the code already. For components we only have e2e, no unit tests yet. I think if we want to invest in this https://testing-library.com/ is kind of standard for the code and https://storybook.js.org/ for the visual.

I will look at writing some for `replaceEntityNodesWithValues` as that's just code.

PR #1305, opened 2019-09-23
feat: Support import of .dialog HTTP actions

tpmsr:
This PR allows .dialog files containing HTTP actions to be imported.  It creates a placeholder API action for any `Microsoft.HttpRequest` step found in the .dialog.

NOTA BENE : this CL assumes the presence of a `responseFields` property in the `OBIDialog`; this is not currently part of the OBI BotBuilder spec.

Apologies for the whitespace changes; many of these were introduced by VSCode's file auto-formatting.


src/Utils/obiDialogParser.ts
```diff
@@ -181,39 +207,89 @@ export class ObiDialogParser {
...
+     * The .dialog file is expected to have a field `responseFields` that enumerates the top-level
+     * output parameters of the response object.  Note that as of 2019.09, this field is specific
+     * to ConversationLearner and is not part of the OBI spec.
+     */
+    private async createActionFromHttpRequest(step: OBIDialog, nextStep: OBIDialog | undefined):
+        Promise<CLM.TrainScorerStep> {
+        if (!step.url) {
+            throw new Error('HTTP requests require url')
+        }
+        const isTerminal = (!nextStep || nextStep.$type === OBIStepType.TEXT_INPUT ||
```
tpmsr: @larsliden does this logic look correct ?

src/Utils/obiDialogParser.ts
```diff
@@ -71,58 +92,53 @@ export class ObiDialogParser {
...
+                        trainDialogs = [...trainDialogs, ...childDialogs]
                     }
                 }
             }
         }
         let trainRound: CLM.TrainRound | undefined
         if (obiDialog.steps) {
-            for (const step of obiDialog.steps) {
-                if (typeof step === "string") {
+            for (let i = 0; i < obiDialog.steps.length; i = i + 1) {
```
mattmazzola: Seems like you only changed this to get the index. Would prefer `for (const [i, step] of obiDialog.steps.entries())`.

src/Utils/obiDialogParser.ts
mattmazzola: Thanks for making it `HttpRequest` instead of `HTTPRequest`

src/Utils/obiDialogParser.ts
```diff
@@ -181,39 +207,89 @@ export class ObiDialogParser {
...
-    
+
+    private async getScorerStepFromActivity(prompt: string): Promise<CLM.TrainScorerStep> {
         const parsedActivity = prompt.substring(prompt.indexOf("[") + 1, prompt.lastIndexOf("]")).trim()
         let response = this.composerDialog.lgMap.get(parsedActivity)
         if (!response) {
             // LARS thow error once CCI .dialog transformer has been fixed
-            response = { text: "Can't Parse LG", suggestions: []}
-           //throw new Error(`LG name ${prompt} undefined`)
+            response = { text: "Can't Parse LG", suggestions: [] }
```
LarsLiden: Would be good to follow up with Swadheen to understand why we are still getting bad data

src/Utils/obiDialogParser.ts
LarsLiden: Probably more complicated.  This is looking only at the current node to check for a next step, but the next step could actually be coming from a node further down in the tree.  Have to think about the best way to do this.  Might be easiest to keep a cache of actions to create and then walk the train dialogs afterwards to determine terminal state.  I'd suggest not blocking check-in of what you have so far, but let's chat and figure out the best approach

src/Utils/obiUtils.tsx
```diff
@@ -62,12 +62,12 @@ export interface TranscriptActionCall {
     type: string
     actionName: string
     actionInput: TranscriptActionInput[]
-    actionOutput: TranscriptActionOutput[]
+    actionOutput: ActionOutput[]
```
LarsLiden: nit - I might name this ObiActionOutput to namespace it / give context.  As it's being exported now "ActionOutput" seems a big generic.

src/types/obiTypes.ts
```diff
@@ -122,6 +122,11 @@ export interface OBIDialog {
      * The property to set the value of
      */
     property?: string;
+    /**
+     * If the result is expected to be structured data such as a JSON object, this holds
+     * the names of the fields or properties in that response object.
+     */
+    responseFields?: string[]
```
LarsLiden: nit - Might not here that this is an added field as the comment at top of file says that it's generated.  If it is re-generated at some point, we'll lose this field

src/Utils/obiDialogParser.ts
tpmsr: Nifty.  Done.

src/Utils/obiDialogParser.ts
```diff
@@ -52,13 +73,13 @@ export class ObiDialogParser {
...
-    
+
         const mainDialog = this.composerDialog.dialogs.find(d => d.$id === "Entry.main")
         if (!mainDialog) {
             return null
         }
-    
-        return await this.getTrainDialogsfromOBIDialog(mainDialog)
+
+        return this.getTrainDialogsfromOBIDialog(mainDialog)
```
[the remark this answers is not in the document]
tpmsr: The "await" raises a tslint "no-return-await".  Seems it's unnecessary to "return await" from a function that is itself async https://palantir.github.io/tslint/rules/no-return-await/

src/Utils/obiUtils.tsx
tpmsr: Done.

src/types/obiTypes.ts
tpmsr: Done.

src/Utils/obiDialogParser.ts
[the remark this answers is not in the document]
LarsLiden: Interesting.

PR #1308, opened 2019-09-25
feat: add tag to dialogs from logs

mattmazzola:
- Adds tags to dialogs from logs
- If dialog is merged with existing dialog and has 'from-log' tag remove the tag to avoid pollution primary/existing dialog.

- Fix bug with tags not being saved on dialogs that have potential merge bug only saved as is.

Took long to find this one as it only occurs under special conditions.
If we don't have potential merge, the tags and description are updated correctly.
If we do have merge and do accept merge tags and description are updated
If we do have merge but do NOT accept it, tags and description were lost

Added an extra "editDialog" call to save these. We could put it up further before the merge modal but that adds extra HTTP request in cases it would be obsolete by next merge dialog request.

I'm not sure why test weren't working on normal run, but here is screenshot
image


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -482,7 +483,6 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
             if (this.props.teachSession.dialogMode !== CLM.DialogMode.EndSession) {
 
                 if (save) {
-
                     // If editing an existing train dialog, extract its dialogId
                     const sourceTrainDialogId = this.sourceTrainDialogId()
```
mattmazzola: Won't let me add comment on the lines below, but this is the other location to add the `editTrainDialogThunkAsync` call.
Would call it every time instead of only on no merge.

PR #1314, opened 2019-09-26
fix: move constants from modals/index.ts to types/const.ts

tpmsr:
This change prevents modal dialogs from being pulled in as dependencies to library code.

Aside from being a rather inappropriate dependency, one of the modals being pulled in was causing errors in jest.

test_fail

^^ note that prior to this change, `AddButtonScore.tsx` is being pulled into `obiUtils.test.ts`


tpmsr:
WIP - this change builds, but I need to update references in the other modals.

tpmsr:
CircleCI builds, submitting!

PR #1320, opened 2019-10-02
feat: n-way ratings and comparison of transcripts

LarsLiden:
image


src/components/modals/TranscriptRatings.tsx
```diff
@@ -16,8 +16,8 @@ import './TranscriptRatings.css'
 interface ComponentState {
     ratePivot: string | undefined
     sourceRankMap: Map<string, RankCount[]>
-    unRatableMap: Map<string, number>    // Number conversations missing transcript for comparison
-    notRatedMap: Map<string, number>    // Number conversations that haven't been rated
+    unRatableMap: Map<string, string[]>    // Number conversations missing transcript for comparison <sourceName, conversationId[]>
```
mattmazzola: If the type of the keys is still string, what is the reason for using `Map` here?

I think the major benefit of using maps is when you need keys of non string type. If only string I wonder if you can use regular object for simpler operations.  I've been meaning to clean up the other instances where we use maps unnecessarily.

Can type as:
```
{ [key: string]: string[] }
OR
Record<string, string[]>
```

src/routes/Apps/App/Testing.tsx
```diff
@@ -535,6 +550,12 @@ class Testing extends React.Component<Props, ComponentState> {
                         </OF.PivotItem>
                     </OF.Pivot>
                 </div>
+                <ConfirmCancelModal
+                    open={this.state.isConfirmClearModalOpen}
```
mattmazzola: What is the reason for the name mismatch?
The modal is `ConfirmCancelModal` but the boolean is `isConfirmClearModalOpen`.
Is there difference between semantics of Cancel vs Clear here?
It would expect `isConfirmCancelModalOpen`

src/components/modals/TranscriptRatings.tsx
LarsLiden: Yup.  Good call.  This was in intermediate step.  They will be <string, string[]> of conversationIds

src/routes/Apps/App/Testing.tsx
LarsLiden: The component name is "ConfirmCancelModel" (used in a lot of places).   The confirmation itself is a warning when the user want to clear the results (i.e. a specific instance of this modal)

PR #1321, opened 2019-10-02
feat: Add ability to mark actions as "entry node"

tpmsr:
Adds a checkbox (visibility protected by flag) to set the "isEntryNode" property on actions.

When this is set, the server should generate out-of-domain utterances that trigger the selected action.

Verified on local client that the checkbox state is persisted when updating actions.

**Notes** :
This is in the UI because it would not be straightforward to drive this from .dialog import.  The .dialog import creates a set of TrainDialogs, representing the possible walks through the dialog tree.  Actions are either created by the user while reviewing the imported dialogs, or are created by selecting the "auto action match" option during import.  In either case, to support automation of this, we'd need to set some flag in the TrainDialog rounds to suggest that the created action should be flagged as an entry node.

The UI does not yet verify that one and only one action has this set.  Forthcoming in separate PR; sending this out now to verify e2e functionality with server, and unblock Swadheen's testing.


tpmsr:
entry_node

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -281,6 +281,7 @@ interface ComponentState {
     secondarySlateValuesMap: SlateValueMap
     isTerminal: boolean
     reprompt: boolean
+    entryNode: boolean
```
mattmazzola: It would be better to the name of property on action model. Like `isTerminal` above, likewise for new `isEntryNode`
There's general practice of boolean variables being named as yes/no questions that can be answered to imply boolean.  `EntryNode` sounds like it could be of type `Node` but we don't have that concept.

LarsLiden:
It’s not any input that is doesn’t understand, only on the first turn.  May want to adjust the verbiage

src/react-intl-messages.ts
```diff
@@ -653,6 +657,7 @@ export default {
         [FM.ACTIONCREATOREDITOR_TRAINDIALOGSBUTTON_TEXT]: 'Train Dialogs',
         [FM.ACTIONCREATOREDITOR_CHECKBOX_TERMINAL_LABEL]: 'Wait for Response?',
         [FM.ACTIONCREATOREDITOR_CHECKBOX_REPROMPT_LABEL]: 'Reprompt on unexected user input',
+        [FM.ACTIONCREATOREDITOR_CHECKBOX_ENTRY_NODE_LABEL]: 'Direct unknown user utterances to this action',
```
mattmazzola: Still think it's weird to mix "input" and "utterance" in same product. Because they're named different people could assume they refer to different things but they are the interchangeable. I prefer "input" as it seems like a more layman term and doesn't imply audio/voice like utterance does but I supposed we'rd already inconsistent on this won't matter.

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -1868,7 +1897,8 @@ const mapStateToProps = (state: State, ownProps: any) => {
         actions: state.actions,
         trainDialogs: state.trainDialogs,
         botInfo: state.bot.botInfo,
-        browserId: state.bot.browserId
+        browserId: state.bot.browserId,
+        settings: state.settings
```
mattmazzola: nit: prefer trailing commas on everything but array items

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -1130,15 +1144,19 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                 ? false
                 : this.state.isTerminal
 
-        // Reprompt only allowed on CARD and TEXT actions
+        // Reprompt and Entry node only allowed on CARD and TEXT actions
         const reprompt = (actionTypeOption.key === CLM.ActionTypes.CARD || actionTypeOption.key === CLM.ActionTypes.TEXT)
             ? this.state.reprompt
             : false
+        const entryNode = (actionTypeOption.key === CLM.ActionTypes.CARD || actionTypeOption.key === CLM.ActionTypes.TEXT)
+        ? this.state.entryNode
```
mattmazzola: nit: formatting

tpmsr:
@lars - done

mattmazzola:
What was proposed PR rules for comments? I thought it was to at least respond to the PR comments before merge even if you don't made suggested change. Helps people understand that you saw it and know why you thought it was not needed and resolve before?

tpmsr:
Matt - my bad, somehow I saw your approve but not your comments.  Will address this morning.

src/components/modals/ActionCreatorEditor.tsx
tpmsr: Updating in new PR

src/react-intl-messages.ts
tpmsr: Updating in new PR.  Note that there are other existing strings that reference "utterance", not modifying those.

src/components/modals/ActionCreatorEditor.tsx
tpmsr: Not sure I follow your comment ?

src/components/modals/ActionCreatorEditor.tsx
tpmsr: Updating in new PR

tpmsr:
@mattmazzola - followup review @ https://github.com/microsoft/ConversationLearner-UI/pull/1323

PR #1322, opened 2019-10-03
fix: store newly-created entities and actions in app memory

tpmsr:
During .dialog import, we check to see if a given action or entity has already been created by looking them up in app memory (TrainDialog's `actions` or `entities` props).  Invoking the `create...ThunkAsync` does not update props while importing a set of dialogs; accordingly, without this change, the import will try to create the same action/entity multiple times, if the same action/entity is referenced in multiple dialog steps.

Tested via import of the product key intent v4 files.

### DIFF
--- a/src/Utils/dialogEditing.ts
+++ b/src/Utils/dialogEditing.ts
@@ -469,8 +469,10 @@ export async function onEditTeach(
     await editHandler(trainDialog, selectedActivity, args)
 }
 
-// Returns placeholder if it exists, otherwise creates it if given creation action.
-// Will add the new action to `actions` if one is created.
+/**
+ * Returns placeholder if it exists, otherwise creates it if given creation action.
+ * Will add the new action to `actions` if one is created.
+ */
 export async function getOrCreatePlaceholderAPIAction(
     appId: string,
     placeholderName: string | "",
@@ -490,9 +492,9 @@ export async function getOrCreatePlaceholderAPIAction(
         placeholder = actions.filter(a => CLM.ActionBase.isPlaceholderAPI(a))
             .map(aa => new CLM.ApiAction(aa))
             .find(aaa => aaa.name === placeholderName)
-        if (placeholder) {
-            return placeholder
-        }
+    }
+    if (placeholder) {
+        return placeholder
     }
 
     // If create action is available create a new action
--- a/src/Utils/dialogEditing.ts
+++ b/src/Utils/dialogEditing.ts
@@ -469,7 +469,8 @@ export async function onEditTeach(
     await editHandler(trainDialog, selectedActivity, args)
 }
 
-// Returns placeholder if it exists, otherwise creates it if given creation action
+// Returns placeholder if it exists, otherwise creates it if given creation action.
+// Will add the new action to `actions` if one is created.
 export async function getOrCreatePlaceholderAPIAction(
     appId: string,
     placeholderName: string | "",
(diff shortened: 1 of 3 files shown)


src/Utils/obiUtils.tsx
```diff
@@ -653,6 +653,9 @@ export async function importActionOutput(
             if (!entityId) {
                 throw new Error("Invalid Entity Definition")
             }
+            // Record the created entity in the input entity list.
+            newEntity.entityId = entityId
+            entities.push(newEntity)
```
LarsLiden: Hmm...  I think it would be better to create a local list rather than push these on to the passed list.  That could have unintended side effects

src/Utils/dialogEditing.ts
```diff
@@ -506,7 +506,7 @@ export async function getPlaceholderAPIAction(
         if (!newAction) {
             throw new Error("Failed to create placeholder API")
         }
-
+        actions.push(newAction)
```
LarsLiden: I think it would be better to create a local list rather than push these on to the passed list.  That could have unintended side effects

mattmazzola: Nice to see preference for immutability now. I also agree that it's best to use pure functions and avoid mutations when possible. This is part of philosophy of react, redux, and declarative based systems.

tpmsr: Would taking a copy of the actions and entities in the ctor be what you had in mind ?

eg
        this.actions = actions
=>
        this.actions = [...actions]

tpmsr: To Matt's point, the only place we know whether or not a given entity or action has been created is in the utility functions that create them.  As currently implemented, we can't tell from the return values whether or not the element being returned is new or memoized.

mattmazzola: Per conversation, if we're going to keep the mutation since alternatives are messy, then would rename the function. The name `getPlaceholderAPIAction` doesn't imply it also modifies list, perhaps `addPlaceholderAPIAction`

tpmsr: @mattmazzola good suggestion, renaming to getOrCreate...

mattmazzola: `getOrCreate` doesn't imply the `actions.push(newAction)` though.  Sign of methods doing too many things when they are hard to name.  This point is caller should be aware that it modifies list.  

It didn't realize this can also return `null` so it's not even a getter... ugh. Do whatever is easier then cause there is no good solution.

tpmsr: Ok - updating comments and merging.

tpmsr: @mattmazzola @LarsLiden - you good with these changes ?

mattmazzola: I suppose a reluctant "yes". I don't like it as name is still misleading and function still mutates it's argument which is likely to cause issues later. Yea, there is a comment but that doesn't make it much better. That being said it doesn't seem like there are any easy solutions because there are many problems surrounding so fixing this wouldn't help much.  This is all part of problem of bad code getting committed. Your task is seemingly small and it would seem unfair to require fixing all this so you go ahead just do the small change. Enough of these add up, and here we are...

tpmsr: Alright; we can chat process more tomorrow, I'll go ahead and merge this to unblock Swadheen's testing.  Thanks!

PR #1325, opened 2019-10-04
feat: aligned webchats

LarsLiden:
compare and ratings view are aligned by adding padding
view scroll together more smoothly
a few other improvements

### DIFF
--- a/src/components/Webchat.tsx
+++ b/src/components/Webchat.tsx
@@ -23,6 +23,8 @@ export function renderActivity(
     renderSelected: ((activity: Activity) => JSX.Element | null) | null,
     editType: EditDialogType,
     shouldRenderHighlight: boolean,
+    padding?: number,
+    hidden?: boolean,
 ): JSX.Element {
 
     const timeLine = <span> {activityProps.fromMe ? "User" : "Bot"}</span>;
@@ -60,19 +62,31 @@ export function renderActivity(
         wrapperClassName += ` wc-message-selected`
     }
 
+    const baseStyle = {}
+    if (padding) {
+        // User can pass in padding to align activities
+        baseStyle['paddingBottom'] = padding
+    }
+    if (hidden) {
+        baseStyle['visibility'] = 'hidden'
+    }
+
     return (
         <div
             data-activity-id={activityProps.activity.id}
             className={wrapperClassName}
             onClick={activityProps.onClickActivity}
             role="button"
+            style={baseStyle}
         >
             <div
                 className={`wc-message wc-message-from-${who} ${messageColor} ${messageBorder}`}
                 ref={div => setRef(div)}
                 data-testid="web-chat-utterances"
             >
-                <div className='wc-message-content'>
+                <div 
+                    className='wc-message-content'
+                >
                     <svg className={`wc-message-callout ${messageFillColor}`}>
                         <path className="point-left" d="m0,6 l6 6 v-12 z" />
                         <path className="point-right" d="m6,6 l-6 6 v-12 z" />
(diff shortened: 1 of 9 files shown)


LarsLiden:
Going to go ahead and merge so I can get this to Swadheen.   Please still review. Thanks.

src/components/Webchat.tsx
```diff
@@ -23,6 +23,8 @@ export function renderActivity(
     renderSelected: ((activity: Activity) => JSX.Element | null) | null,
     editType: EditDialogType,
     shouldRenderHighlight: boolean,
+    padding?: number,
+    hidden?: boolean,
 ): JSX.Element {
```
mattmazzola: I noticed you are you use `JSX.Element` instead of `React.ReactNode` and wondered if this is intentional or not? I thought `React.ReactNode` was preferred when possible.

src/components/Webchat.tsx
```diff
@@ -278,6 +293,8 @@ export interface ReceivedProps {
     onPostActivity: (a: Activity) => void,
     onScrollChange?: (position: number) => void,
     renderActivity?: (props: BotChat.WrappedActivityProps, children: React.ReactNode, setRef: (div: HTMLDivElement | null) => void) => (JSX.Element | null)
+    // Callback for rendered height of an activity
+    onActivityHeight?: (index: number, height: number) => void
```
mattmazzola: When does this get called?

src/components/modals/CompareDialogsModal.tsx
```diff
@@ -133,8 +180,8 @@ class CompareDialogsModal extends React.Component<Props, ComponentState> {
         }
 
         let logDialogId: string | undefined
-        const activityMap = new Map<string, BB.Activity[]>()
-        const rankMap = new Map<string, number | undefined>()
+        const activityMap = {}
```
mattmazzola: I'm supersized this worked. Isn't it supposed to be typed as `{[key: string]: BB.Activity[]}`  and same for other.
Assigning blank object seems weird. The type information on map provided more guarantees

src/components/modals/RateDialogsModal.tsx
```diff
@@ -28,26 +29,35 @@ interface ComponentState {
     ratingPair: Test.RatingPair | undefined
     resultIndex: number
     webchatKey: number,
-    activities1: BotChat.Activity[] | undefined
-    activities2: BotChat.Activity[] | undefined
+    activities1: BB.Activity[] | undefined
```
mattmazzola: Why do we sometimes use activity from the custom webchat and somtimes bot builder? I would expect it to always come from bot builder?

src/components/modals/RateDialogsModal.tsx
```diff
@@ -237,9 +315,49 @@ class RateDialogsModal extends React.Component<Props, ComponentState> {
...
+                        ac.sourceName === sourceName && ac.index === index)!
+                    activityHeight.height = height
+                    return {activityHeights: prevState.activityHeights}
+                })
+            } 
+        }
+    }
+
+    getRenderData(): RenderData[] {
+
```
mattmazzola: nit: formatting

src/components/Webchat.tsx
LarsLiden: when height changes.  update comment

src/components/modals/CompareDialogsModal.tsx
LarsLiden: How else would you initialize an empty instance?

src/components/modals/RateDialogsModal.tsx
LarsLiden: what formatting?

src/components/modals/RateDialogsModal.tsx
LarsLiden: Working on switching all Activity refs over to BB.

src/components/Webchat.tsx
LarsLiden: done

src/components/modals/RateDialogsModal.tsx
mattmazzola: you have extra blank lines, looks out of place. If we want to make a rule that all methods have blank lines under them we should do it to all of them, other wise we don't do it to any of them. People need to be able to recognize patterns otherwise they just do whatever

LarsLiden: Ah, I see.   Got confused as it's already fixed on my local copy

PR #1326, opened 2019-10-04
fix: allow import of .dialog's with commented json

tpmsr:
`JSON.parse()` fails on import with JSON that contains comments.

This change strips the comments from .dialog JSON files prior to parsing via strip-json-comments, which handles `//`, `/* */`, and `#` style comments.

Tested locally with `//` files.


tpmsr:
Hm, somehow this picked up an unsubmitted change from Lars.  Closing and regenerating.

PR #1327, opened 2019-10-04
fix: allow import of .dialog's with JSON comments

tpmsr:
JSON.parse() fails on import with JSON that contains comments.

This change strips the comments from .dialog JSON files prior to parsing via strip-json-comments, which handles `//`, `/* */`, and `#` style comments.

Tested locally with `//` files.

### DIFF
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -6,6 +6,7 @@ import * as CLM from '@conversationlearner/models'
 import * as DialogEditing from './dialogEditing'
 import * as OBIUtils from './obiUtils'
 import * as Util from './util'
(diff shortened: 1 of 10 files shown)


src/Utils/obiDialogParser.ts
```diff
@@ -6,6 +6,7 @@ import * as CLM from '@conversationlearner/models'
 import * as DialogEditing from './dialogEditing'
 import * as OBIUtils from './obiUtils'
 import * as Util from './util'
+import * as stripJson from 'strip-json-comments'
```
mattmazzola: Why did you remove the "comments" part of the import?

`stripJson` now it sounds like it's stripping out json.... from json.  
if you want to shorten name call it `stripComments`

PR #1329, opened 2019-10-08
feat: attach lg name to action 

LarsLiden:
- elminate multple sources of Activity definition
- rename validationSet to testSet
- store transcript with lgName only
- add lgName to action when matched

### DIFF
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -6,7 +6,7 @@ import * as CLM from '@conversationlearner/models'
 import * as DialogEditing from './dialogEditing'
 import * as OBIUtils from './obiUtils'
 import * as Util from './util'
-import { OBIDialog } from '../types/obiTypes'
+import * as OBITypes from '../types/obiTypes'
 
 enum OBIStepType {
     BEGIN_DIALOG = "Microsoft.BeginDialog",
@@ -20,11 +20,21 @@ enum OBIRuleType {
     INTENT_RULE = "Microsoft.IntentRule"
 }
 
+export interface ObiDialogParserResult {
+    luMap: Map<string, string[]>
+    lgItems: CLM.LGItem[],
+    trainDialogs: CLM.TrainDialog[]
+    warnings: string[]
+}
+
 export class ObiDialogParser {
     private app: CLM.AppBase
-    private composerDialog: OBIUtils.ComposerDialog
+    //private composerDialog: OBIUtils.ComposerDialog
     private actions: CLM.ActionBase[] = []
     private entities: CLM.EntityBase[] = []
+    private dialogs: OBITypes.OBIDialog[]
+    private luMap: Map<string, string[]>
+    private warnings: string[]
     private createActionThunkAsync: (appId: string, action: CLM.ActionBase) => Promise<CLM.ActionBase | null>
     private createEntityThunkAsync: (appId: string, entity: CLM.EntityBase) => Promise<CLM.EntityBase | null>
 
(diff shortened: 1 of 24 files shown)


src/Utils/obiDialogParser.ts
```diff
@@ -21,11 +21,21 @@ enum OBIRuleType {
     INTENT_RULE = "Microsoft.IntentRule"
 }
 
+export interface ObiDialogParserResult {
+    luMap: Map<string, string[]>
```
mattmazzola: You seemed to be avoiding string maps in other PRs but chose to use here even though key is still `string`.
What was reason for this one?

src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -150,14 +150,15 @@ const getDialogKey = (trainDialog: OF.IObjectWithKey) => (trainDialog as CLM.Tra
...
-    activityHistory: Activity[]
+    activityHistory: BB.Activity[]
     lastAction: CLM.ActionBase | null
     isTeachDialogModalOpen: boolean
     isEditDialogModalOpen: boolean
     isTranscriptImportOpen: boolean
     isImportWaitModalOpen: boolean
     importIndex: number | undefined
     importedTrainDialogs: CLM.TrainDialog[] | undefined
+    importedLGItems: CLM.LGItem[] | undefined
```
mattmazzola: Hmm, I thought you had changed preferences after seeing the `lgName` addition, but now I see this and realize it was a trick. What is the logic for capitalization?
We have mixes of `LGItem` `lgItem` and others. Whatever the pattern is, it should be clear enough that other people can repeat it.

LarsLiden: In my opinion, variable names shouldn't start with caps, thus "lgItems".  However if it's in the middle leave it caps.  'importedlgItems" is less readable than "importedLGItems".

mattmazzola: Well would be `importedLgItems` it's `importedlgItems` which avoids the i L issue, but that's only specific because of this word anyways. I don't agree it's more readable but at least I understand why they're different now.

LarsLiden: My other thought is that LG is an abbreviation itself, so fro me Lg reads strange

src/Utils/obiDialogParser.ts
mattmazzola: Well given it's merged already suppose it doesn't matter. But still interested in why they are treated differently

LarsLiden: My understanding was that we agreed on doing post-merge changes for stuff the is time sensitive.    In any case, in answer to your comment, laziness. :)   I've been gradually switching stuff over and was editing the lg part.  LU not really used yet, but I'll make that change too.

mattmazzola: Hmm. I suppose I suspected this but wanted to believe otherwise.  Makes me not want to do code reviews if comments are skipped out of laziness. 

> I've been gradually switching stuff over 

This is just more laziness. It's easier for you to makes changes where you're currently working instead of everywhere but it's intentionally introducing inconsistency or pattern that contradicts the old. You lowered your cost of changes cost was then multiplied by everyone else who sees it.  Just leads to confusion. We wouldn't be having this conversation if it wasn't there so now it's taking more time.

LarsLiden:
Let’s take this conversation off-line

tpmsr:
Something in this PR broke action matching; .dialogs with HTTP requests are now coming up as "can't find action $GUID" (referencing the wrong GUID).
missing action

It works if I sync to the commit prior to this PR.  The attached zip has a tiny dialog that allows repro.
[action-dialog.zip](https://github.com/microsoft/ConversationLearner-UI/files/3709909/action-dialog.zip)

tpmsr:
Ok, red herring - the action import is actually failing non-deterministically.  Lars and I walked through it this morning and he identified the root case; I incorporated the fix into https://github.com/microsoft/ConversationLearner-UI/pull/1331

PR #1330, opened 2019-10-09
feat: import warnings - group import properties in train dialogs

LarsLiden:


tpmsr:
`needs-review` or `fyi` ?

LarsLiden:
Adding tag now.  Thx for reminder

src/Utils/obiDialogParser.ts
```diff
@@ -22,7 +22,7 @@ enum OBIRuleType {
 }
 
 export interface ObiDialogParserResult {
-    luMap: Map<string, string[]>
+    luMap: { [key: string]: string[] }
```
tpmsr: Just curious, what's the benefit of changing from a `Map` to a dict here ?

src/routes/Apps/App/TrainDialogs.css
```diff
@@ -6,4 +6,9 @@
     display: grid;
     grid-template-columns: auto max-content;
     grid-gap: 1em;
+}
+
+.cl-traindialogs-import-warning {
+    height: 200px;
+    overflow-y: auto;
 }
```
tpmsr: nit: Add a final newline while you're in this file

src/Utils/obiDialogParser.ts
LarsLiden: Matt's suggestion.

tpmsr: Ah, from the previous review.  Got it.

@mattmazzola - is this a general best practice in JS/TS when the key is a string ?  Any details on why ?

mattmazzola: Well I would consider that a lazy answer. Glad that you changed because the suggestion came from me but in future I would hope you understand the reasoning of the suggestion and why  instead of just blindly changing code without understanding. That likely leads to more bugs.
It also doesn't answer the question "what's the benefit". 

Benefits are cleaner/simpler API and the follows expectation of reader.  You use `Maps` when they are required.  Such as when keys are not strings, or need to take advantage of some optimizations that data structure provides like the `size` property instead of `Object.keys()`.
If you use a `Map` reader thinks it was needed and then searching for case where it was required when there is no case. There is `Record<K,V>` type for similar syntax of typing and object that way also.

PR #1331, opened 2019-10-09
feat: OBI dialog import as a tree walk

tpmsr:
This PR breaks apart OBI .dialog file into discrete steps : 

1. Generation of the dialog tree structure
2. Building TrainDialogs from each walk of the dialog tree

This is a precursor to adding support for SwitchCondition statements, which are complicated in that they generate dialog branches (similar to `rule` elements) but also can contain content (similar to `step` elements).

### DIFF
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -7,7 +7,7 @@ import * as DialogEditing from './dialogEditing'
 import * as OBIUtils from './obiUtils'
 import * as Util from './util'
 import * as OBITypes from '../types/obiTypes'
-import * as stripJson from 'strip-json-comments'
+import * as stripJsonComments from 'strip-json-comments'
 
 enum OBIStepType {
     BEGIN_DIALOG = "Microsoft.BeginDialog",
@@ -21,6 +21,17 @@ enum OBIRuleType {
     INTENT_RULE = "Microsoft.IntentRule"
 }
 
+// A node in a dialog tree.
+class ObiDialogNode {
+    readonly dialog: OBITypes.OBIDialog
+    intent?: string
+    children: ObiDialogNode[]
+    constructor(dialog: OBITypes.OBIDialog) {
+        this.dialog = dialog
+        this.children = []
+    }
+}
+
 export interface ObiDialogParserResult {
     luMap: Map<string, string[]>
     lgItems: CLM.LGItem[],
@@ -30,10 +41,9 @@ export interface ObiDialogParserResult {
 
 export class ObiDialogParser {
     private app: CLM.AppBase
-    //private composerDialog: OBIUtils.ComposerDialog
     private actions: CLM.ActionBase[] = []
     private entities: CLM.EntityBase[] = []
-    private dialogs: OBITypes.OBIDialog[]
+    private dialogs: Map<string, OBITypes.OBIDialog>
     private luMap: Map<string, string[]>
     private warnings: string[]
     private createActionThunkAsync: (appId: string, action: CLM.ActionBase) => Promise<CLM.ActionBase | null>
@@ -53,24 +63,20 @@ export class ObiDialogParser {
         this.createEntityThunkAsync = createEntityThunkAsync
     }
 
-    async parse(files: File[]): Promise<ObiDialogParserResult> {
-
-        const lgItems: CLM.LGItem[] = []
-        this.luMap = new Map()
-        this.dialogs = []
-        this.warnings = []
-
+    // Reads input files; packs data into dialog / LU / LG maps according to file extensions.
+    async readDialogFiles(files: File[], dialogs: Map<string, OBITypes.OBIDialog>, luMap: Map<string, string[]>,
+        lgItems: CLM.LGItem[]) {
         for (const file of files) {
             if (file.name.endsWith('.dialog')) {
                 const fileText = await Util.readFileAsync(file)
-                const obiDialog: OBITypes.OBIDialog = JSON.parse(stripJson(fileText))
+                const obiDialog: OBITypes.OBIDialog = JSON.parse(stripJsonComments(fileText))
                 // Set name, removing suffix
                 obiDialog.$id = this.removeSuffix(file.name)
-                this.dialogs.push(obiDialog)
+                dialogs.set(obiDialog.$id, obiDialog)
             }
             else if (file.name.endsWith('.lu')) {
                 const fileText = await Util.readFileAsync(file)
-                this.addToLUMap(fileText, this.luMap)
+                this.addToLUMap(fileText, luMap)
             }
             else if (file.name.endsWith('.lg')) {
                 const fileText = await Util.readFileAsync(file)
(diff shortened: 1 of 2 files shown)


src/Utils/obiDialogParser.ts
```diff
@@ -112,51 +123,141 @@ export class ObiDialogParser {
...
+        return node
+    }
 
-                    const textVariations = this.getTextVariations(rule.intent!)
-                    const extractorStep: CLM.TrainExtractorStep = {
-                        textVariations: textVariations
-                    }
-                    if (!rule.steps) {
+    // Collects dialog tree nodes from `Microsoft.IntentRule` elements in the dialog `rules` section.
+    private async collectDialogRuleChildren(node: ObiDialogNode, rules: OBITypes.MicrosoftIRule[]) {
```
mattmazzola: Don't have much context on the OBI stuff but it's weird for recursive functions (`collectDialogRuleChildren`, `collectDialogStepChildren`) to modify their inputs `node.children.push(child)`

I would expect given a node and rules to return the children, and then next statement can choose to add the children to the node. Similar to other conversation about trying for pure functions, although I assume this codebase past that point so maybe it's too much to change.

src/Utils/obiDialogParser.ts
```diff
@@ -30,10 +41,10 @@ export interface ObiDialogParserResult {
 
 export class ObiDialogParser {
     private app: CLM.AppBase
-    //private composerDialog: OBIUtils.ComposerDialog
     private actions: CLM.ActionBase[] = []
     private entities: CLM.EntityBase[] = []
-    private dialogs: OBITypes.OBIDialog[]
+    private dialogs: Map<string, OBITypes.OBIDialog>
```
mattmazzola: Idk if you just following file convention since there are other `Map`s below, but I thought there was just talk on two previous reviews and only using maps when they're required so people know to watch out.

What prevents this from being `Record<string, OBITypes.OBIDialog>` ?

src/Utils/obiDialogParser.ts
tpmsr: Maybe this is more of a Java / C++ pattern, but in my experience it's pretty common to add things to collections via out-parameters in recursive functions that have this kind of structure.  Out-params are more obvious in C++ where you'd have a pointer parameter.

Happy to change if you feel strongly about it.

mattmazzola: Yea, I'm sure it is more common in those languages although we're not writing those languages. "out-parameters" don't exist in javascript and the popularity of those languages peaked while people still largely wrote imperative code and tried to optimized for computation or memory both of which are lesser concerns now. Javascript, in particular React, is much different, more functional and declarative.

Than being said, I can't say it would be better to change and follow patterns of rest of repo because there are such as mix of functional / non-functional stuff that it's not very clear anymore. I don't know what to feel strongly about anymore. I would say leave it unless you also want to change it

src/Utils/obiDialogParser.ts
```diff
@@ -53,52 +64,52 @@ export class ObiDialogParser {
...
 
-    async parse(files: File[]): Promise<ObiDialogParserResult> {
-
-        const lgItems: CLM.LGItem[] = []
-        this.luMap = new Map()
-        this.dialogs = []
-        this.warnings = []
-
+    // Reads input files; packs data into dialog / LU / LG maps according to file extensions.
+    async readDialogFiles(files: File[]) {
```
LarsLiden: should this be private?

src/Utils/obiDialogParser.ts
```diff
@@ -112,51 +123,141 @@ export class ObiDialogParser {
...
-                    const textVariations = this.getTextVariations(rule.intent!)
-                    const extractorStep: CLM.TrainExtractorStep = {
-                        textVariations: textVariations
-                    }
-                    if (!rule.steps) {
+    // Collects dialog tree nodes from `Microsoft.IntentRule` elements in the dialog `rules` section.
+    private async collectDialogRuleChildren(node: ObiDialogNode, rules: OBITypes.MicrosoftIRule[]) {
+        for (const rule of rules) {
+            if (rule.$type !== OBIRuleType.INTENT_RULE) {
+                console.log(`Unhandled OBI rule type: ${rule.$type}`)
```
LarsLiden: Consider passing this back in warning array

src/Utils/obiDialogParser.ts
```diff
@@ -112,51 +123,141 @@ export class ObiDialogParser {
...
+                continue
+            }
+            const intent = rule.intent
+            if (!intent) {
+                throw new Error(`Rule is missing intent property`)
```
LarsLiden: Should this be a throw or a warning?   If a catastrophic failure, then throw

src/Utils/obiDialogParser.ts
```diff
@@ -112,51 +123,141 @@ export class ObiDialogParser {
...
+            }
+            if (!rule.steps) {
+                continue
+            }
+            for (const step of rule.steps) {
+                if (typeof step === "string") {
+                    throw new Error("Unexpected string step")
+                }
+                if (step.$type !== OBIStepType.BEGIN_DIALOG || typeof step.dialog !== "string") {
+                    console.log(`Unhandled OBI Type: ${step.$type}`)
```
LarsLiden: Generally think about whether any of these console logs or throws should be warnings

src/Utils/obiDialogParser.ts
```diff
@@ -191,39 +294,21 @@ export class ObiDialogParser {
...
-                    if (!subDialog) {
-                        throw new Error(`Dialog name ${step.dialog} undefined`)
-                    }
-
-                    const childDialogs = await this.getTrainDialogsfromOBIDialog(subDialog)
-
-                    // Add children to train dialog list
-                    trainDialogs = [...trainDialogs, ...childDialogs]
+                // TODO(thpar) : handle Microsoft.SwitchCondition.
+                case OBIStepType.BEGIN_DIALOG: {
```
LarsLiden: Should add a default case

tpmsr: There is one, see line 304 below

src/Utils/obiDialogParser.ts
tpmsr: Oh, I just realized how I missed your comments on that other PR.  GitHub doesn't show comments on lines that have changed since a PR has been pushed.  That seems like a horrific bug.

I was following the style of the existing code; updated to dict, similar to other var.

src/Utils/obiDialogParser.ts

src/Utils/obiDialogParser.ts

src/Utils/obiDialogParser.ts
tpmsr: Updated most to warnings.  Keeping missing dialog as error.

src/Utils/obiDialogParser.ts
tpmsr: Since `collectDialogNodes()` is recursively reentrant from multiple callsites -- and `node` changes on each call -- I think it's cleaner and easier to follow to modify the node in the current context as is done here.

I have, however, modified `getTrainDialogs()` to return the dialogs rather than adding to the collection parameter.

tpmsr:
I'm getting the dreaded `ERROR: Cannot read property 'labelEntities' of undefined` again, which doesn't seem to be going away with sync + `npm install` and rebuild of either this or the Samples repo.

I'll go ahead and merge this code, and dig into the issue with the team on Monday.

PR #1332, opened 2019-10-10
feat: allow adding conditions to actions

mattmazzola:
Can't be merged until server changes so there is less urgency.

## Adds `ConditionCreatorModal` to create conditions (currently only Enum and Value conditions)
image

image

## Changes to render ValueConditions on actions details list and action scorer list

image
image

### Better visualized with [Video](https://microsoft.sharepoint.com/:v:/t/convsysresearch/EdZpxCMdu3ZLsBvv_YvYw94Bia68jg8eD-mxvMs1PSwXXw?e=1SejZ7)

E2E tests:
image


src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -508,7 +563,7 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
                     expectedEntityTags,
                     negativeEntityTags,
                     requiredEntityTagsFromPayload,
-                    requiredEntityTags,
+                    requiredEntityOrConditionTags: requiredEntityTags,
```
mattmazzola: Looks like there is more renaming to do here. Will change.

src/react-intl-messages.ts
```diff
@@ -972,7 +972,7 @@ export default {
         [FM.TOOLTIP_ACTION_RESPONSE_ROW3]: '"Hi[, $name]"',
 
         [FM.TOOLTIP_ACTION_REQUIRED]: 'Action will not be selected unless Memory contains values for these Entities',
-        [FM.TOOLTIP_ACTION_REQUIRED_TITLE]: 'Required Entities',
+        [FM.TOOLTIP_ACTION_REQUIRED_TITLE]: 'Required Entities or Conditions',
```
mattmazzola: I had thought that we could drop `Entities` wording here. The entity being present is just another condition.

Instead of user having to think entities and conditions are different, the entity being present is just another condition where the entity value can be anything as opposed to conditions where entity value must be an Enum value or must be greater than some number.  Would simply concepts for users and avoid exposure to ML concepts like 'entity' average user likely more commonly understands the context and implications of word "condition" that what it means to require an entity.

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -315,7 +390,7 @@ const initialState: Readonly<ComponentState> = {
     availableConditionalTags: [],
     expectedEntityTags: [],
     requiredEntityTagsFromPayload: [],
-    requiredEntityTags: [],
+    requiredEntityOrConditionTags: [],
     negativeEntityTags: [],
```
LarsLiden: Why not the same on negativeEntityTags?

src/components/modals/ActionScorer.tsx
```diff
@@ -40,6 +40,38 @@ interface IRenderableColumn extends OF.IColumn {
     render: (actionForRender: ActionForRender, component: ActionScorer, index: number) => React.ReactNode
 }
 
+const findNumberFromMemory = (memory: CLM.Memory, isMultivalue: boolean): number | undefined => {
+    if (isMultivalue) {
```
LarsLiden: Is multivalue supported yet or prep for when it is?

src/components/modals/ConditionCreatorModal.tsx
```diff
@@ -0,0 +1,322 @@
...
+                                    <div className="cl-condition-creator__existing-condition"
+                                        data-testid="condition-creator-modal-existing-condition">
+                                        {conditionalTag.name}
+                                    </div>
+
+                                    <OF.DefaultButton
+                                        data-testid="condition-creator-modal-button-use-condition"
+                                        onClick={() => onClickExistingCondition(conditionalTag.condition!)}
+                                    >
+                                        Use Condition
```
LarsLiden: I'm not sure I'm a fan of the big list of "Use Condition" buttons.  Why not make the condition text in to a button?

src/react-intl-messages.ts
LarsLiden: Interesting idea.  Run it by Swadheen?

src/components/modals/ConditionCreatorModal.tsx
mattmazzola: Usability. You want user to know it's clickable and describe what it will do when clicked.
They are familiar with buttons being clickable clickable and it has label to describe what it will do when clicked.

Clickable conditions do neither of those things.

src/components/modals/ActionScorer.tsx
mattmazzola: I didn't know what you meant by "supported"?

Single Value number entities use the "number from label"
Multi Value entities use the "number of labels"

src/components/modals/ActionCreatorEditor.tsx
mattmazzola: Yea, I had left comment saying there is more renaming to do:
https://github.com/microsoft/ConversationLearner-UI/pull/1332#discussion_r333652689

src/react-intl-messages.ts
mattmazzola: @swadheen Thoughts?

src/components/modals/ConditionCreatorModal.tsx
LarsLiden: I'm suggesting make them look like buttons 
image
@swadheen  - what do you think?

src/components/modals/ActionScorer.tsx
LarsLiden: Asking if work to make multi-value conditions work was also done (looks like it was?)

src/components/modals/ConditionCreatorModal.tsx
mattmazzola: That doesn't give any information about what happens when you click the button. You only solved half of the problem

And adding "use condition" in there is too noisy. Buttons should be actions like "Cancel" "Save" "Create" etc. or in this case "Use".

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -315,8 +332,8 @@ const initialState: Readonly<ComponentState> = {
     availableConditionalTags: [],
     expectedEntityTags: [],
     requiredEntityTagsFromPayload: [],
-    requiredEntityTags: [],
-    negativeEntityTags: [],
+    requiredEntityOrConditionTags: [],
+    negativeEntityOrConditionTags: [],
```
tpmsr: nit : alphabetize (just the edits, or the full list)

src/components/modals/ActionScorer.tsx
tpmsr: This function could use a comment explaining the behaviour for various entity value types.

src/components/modals/ActionScorer.tsx
```diff
@@ -40,6 +41,38 @@ interface IRenderableColumn extends OF.IColumn {
...
+    }
+
+    const valueString: string | undefined = memory
+        && memory.entityValues[0]
+        && (memory.entityValues[0].resolution ? true : undefined)
+        && (memory.entityValues[0].resolution as any).value as string
+
+    const value = valueString
+        ? parseInt(valueString)
+        : undefined
```
tpmsr: Should this throw an error on non-numeric values ?

src/components/ActionDetailsList.tsx
```diff
@@ -299,34 +300,36 @@ function renderConditions(entityIds: string[], conditions: CLM.Condition[], allE
...
         const entity = allEntities.find(e => e.entityId === entityId)
-        if (!entity) {
-            elements.push(renderCondition(`Error - Missing Entity ID: ${entityId}`, isRequired))
-        }
-        else {
-            elements.push(renderCondition(entity.entityName, isRequired))
-        }
+        return !entity
+            ? renderCondition(`Error - Missing Entity ID: ${entityId}`, isRequired)
+            : renderCondition(entity.entityName, isRequired)
```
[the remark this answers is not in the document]
mattmazzola: > both branches are equal length

What did you mean by this?

> often clearer to write the ternary as

Don't agree. Maybe for simple constants, like example you have where it was `a ? b : c` but for anything over that believe it's better to separate for readability/comparison/editing.

```typescript
if (condition) {
  doA()
} else {
  doB()
}
```
Transformation to ternary expression preserve form and removes distracting syntax from meaningful content
```typescript
condition
  ? doA()
  : doB()
```

That being said, on second look I can simplify it further and will do that.

```typescript
const name = !entity
  ? `Error - Missing Entity ID: ${entityId}`
  : entity.entityName

return renderCondition(name, isRequired)
```

src/components/modals/ActionScorer.tsx
mattmazzola: Hmm, I didn't think so. Think it's better to be flexible here and take in any type of memory value otherwise we just add more layers.  Wonder what others think.

src/components/modals/ActionScorer.tsx
mattmazzola: To answer both:
1. yes I think multi value work was done (depends the cardinality / number of memory values)
2. will add comment

src/components/ActionDetailsList.tsx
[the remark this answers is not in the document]
tpmsr: What I mean is that checking for the positive condition `return (foo) ? ...` is often preferable to checking for the negative condition `return (!foo) ? ...` if the cases are of similar length / complexity.

IOW I would suggest writing this as

```
        return entity
            : renderCondition(entity.entityName, isRequired)
            ? renderCondition(`Error - Missing Entity ID: ${entityId}`, isRequired)
```
or in your refactored version

```
const name = entity
  ? entity.entityName
  : `Error - Missing Entity ID: ${entityId}`
```

It's a super minor point.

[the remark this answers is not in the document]
mattmazzola: Oh I see. It was about using positive assertions `entity` vs `!entity`, not cleanliness of `a ? b : c`.

src/Utils/actionCondition.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    let conditionalTag: IConditionalTag
+    if (entity.entityType === CLM.EntityType.ENUM) {
+        if (!entity.enumValues) {
+            throw new Error(`Condition refers to Entity without Enums ${entity.entityName}`)
+        }
+
+        const enumValueId = condition.valueId
+        if (!enumValueId) {
+            throw new Error(`Condition refers to enum entity, but condition did not have enum value id.`)
```
tpmsr: add entity name to error msg

src/Utils/actionCondition.ts
```diff
@@ -0,0 +1,123 @@
...
+        }
+        const enumValue = entity.enumValues.find(e => e.enumValueId === enumValueId)
+        if (!enumValue) {
+            throw new Error(`Condition refers to non-existent EnumValue: ${enumValueId}`)
```
tpmsr: add entity name to error msg

src/Utils/actionCondition.ts
```diff
@@ -0,0 +1,123 @@
...
+        && (memory.entityValues[0].resolution as any).value as string
+
+    const value = valueString
+        ? parseInt(valueString)
+        : undefined
+
+    return value
+}
+
+export const isValueConditionTrue = (condition: CLM.Condition, numberValue: number): boolean => {
```
tpmsr: This method feels like it belongs in Models, where the conditions are applied.

src/react-intl-messages.ts
mattmazzola: Pinging @swadheen again..,

swadheen: Sounds good.. Thanks!

src/Utils/actionCondition.ts
mattmazzola: Saw your other PR on models repo and reminded me of this comment. I would say Models was named that way because it was originally intended to only be models. As a set of types/interfaces shared by the UI and SDK to ensure they transferred same objects over HTTP.  Since types and interface don't translate to any material code this made the repo extremely light weight. There has been more and more stuff put in there and it's starting to become something other than models, but in general it was only moved if it was required to be the same and actually used in both places.

That being said, there is a different `isValueConditionTrue` in SDK that might be appropriate to consolidate.

tpmsr: SDK also SGTM

PR #1333, opened 2019-10-11
feat: transcript ratings are transitive

LarsLiden:
decreases number of required user ratings

### DIFF
--- a/src/types/TestObject.test.ts
+++ b/src/types/TestObject.test.ts
@@ -1,0 +1,212 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import * as Test from './TestObjects'
+
+describe('TestObject', () => {
+
+    const ALPHA = "alpha"
+    const BETA = "beta"
+    const CHARLIE = "charlie"
+    const conversationId = "1234" 
+
+    const makePair = (sourceA: string, sourceB: string, result: Test.RatingResult): Test.RatingPair => {
+        const sourceNames = [sourceA, sourceB].sort()
+        return {
+            conversationId,
+            sourceNames: [sourceNames[0], sourceNames[1]],
+            result
+        }
+    }
+
+    const addPair = (testSet: Test.TestSet, sourceA: string, sourceB: string, result: Test.RatingResult): void => {
+        testSet.ratingPairs.push(makePair(sourceA, sourceB, result))
+    }
+
+    describe('setFirstIsBetter', () => {
+
+        test('alphaFirst', () => {
+
+            const testSet = Test.TestSet.Create()
+            addPair(testSet, ALPHA, BETA, Test.RatingResult.UNKNOWN)
+            testSet.setFirstIsBetter(ALPHA, BETA, conversationId)
+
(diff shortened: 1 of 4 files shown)


src/types/TestObject.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    const ALPHA = "alpha"
+    const BETA = "beta"
+    const CHARLIE = "charlie"
+    const conversationId = "1234" 
+
+    const makePair = (sourceA: string, sourceB: string, result: Test.RatingResult): Test.RatingPair => {
+        const sourceNames = [sourceA, sourceB].sort()
+        return {
+            conversationId,
+            sourceNames: [sourceNames[0], sourceNames[1]],
```
mattmazzola: What is the purpose if this line?
It seems equivalent to `sourceNames`

src/types/TestObject.test.ts
```diff
@@ -0,0 +1,212 @@
...
+            result
+        }
+    }
+
+    const addPair = (testSet: Test.TestSet, sourceA: string, sourceB: string, result: Test.RatingResult): void => {
```
mattmazzola: The function seems unnecessary.  It doesn't offer more value then the line itself.

src/types/TestObjects.ts
```diff
@@ -444,6 +501,33 @@ export class TestSet {
...
+                return ratingPair.sourceNames[1]
+            }
+        }
+        return null
+    }
+
+    // Update rating pairs to indicate one source is better than another
+    setFirstIsBetter(better: string, worse: string, conversationId: string) {
+        const ratingPair = this.getRatingPair(better, worse, conversationId)
+        if (ratingPair) {
```
mattmazzola: Is this function named as you intended? the `FirstIsBetter` doesn't seem to make sense. It's not checking which is first or second, it's checking if string is better or worse and then setting the order first or second.

It looks like it's setting `FIRST` on the `better` string.  Seems like `setFirstOnBetter` or `setFirstIfBetter`

src/types/TestObject.test.ts
LarsLiden: sourceNames needs to be alphabetically sorted

LarsLiden: Also it is typed as an array of length 2 and result from sort is not

src/types/TestObject.test.ts
LarsLiden: I think it makes the code easier to read

src/types/TestObject.test.ts
mattmazzola: > sourceNames needs to be alphabetically sorted. 
They are sorted by `sort()` fuction not this line that takes them out of array and then puts them back in.

> it is typed as an array of length 2
Then specify this constraint on the `sourceNames` type instead of doing weird manipulation of values to simulate intention.
`const sourceNames = ['a', 'b'].sort() as [string, string]`

src/types/TestObjects.ts
LarsLiden: RatingPair assumes the sourceNames are in alphabetical order.  So when you say the first one is better, whether you set FIRST or SECOND depends on how they sort

PR #1334, opened 2019-10-11
fix: bug 2305 save as is fails on merge

LarsLiden:


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -501,6 +501,8 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
                         // If editing an existing Train Dialog, replace existing with the new one
                         if (sourceTrainDialogId) {
                             await ((this.props.trainDialogReplaceThunkAsync(this.props.app.appId, sourceTrainDialogId, newTrainDialog) as any) as Promise<void>)
+                            // Grab the replaced version
+                            newTrainDialog = this.props.trainDialogs.find(td => td.trainDialogId === sourceTrainDialogId)!
```
mattmazzola: Seems like your an expecting the `trainDialogReplaceThunkAsync` to ensure a dialog of that `trainDialogId` is in the `this.props.trainDialogs` collection which is why you added the `!` operator. Would be better to throw error if it's not found then silently fail later.  Probably a common operation around codebase we could fix in other places.

LarsLiden: good point

PR #1335, opened 2019-10-14
fix: handle edge cases for OBI .dialog import

tpmsr:
In OBI .dialog files, some steps in the tree can carry an "intent", which is a dialog sub-tree classification such as `Product_Key_Service_Integration_main` or `End_of_Conversation_main`.  We use this "intent" as a proxy for user utterances in the generated train dialogs :

end_of_conv

This PR fixes 2 edge cases during .dialog import.

1. Nodes that have an incoming intent but no actions other than redirection to other dialog nodes
2. Nodes that have actions but no incoming intents

**For (1)**, the current node's intent was only used if the node also generated actions; this would cause the intent to be dropped if the only action was to redirect to another dialog node.

To fix this, the intent is carried forward through the recursive handling of the dialog tree if it's not used in the generation of a TrainRound.

**For (2)**, we could have a situation where a node generates a valid round -- eg, it has both an incoming intent and one or more actions -- and then the node redirects to another node that has only action steps.  In this situation, we cannot create a new train round for the action-only node, since there is no new user utterance (intent) being introduced.

To fix this, if a node generates only `ScorerStep`s (and no `ExtractorStep`), those `ScorerStep`s will be appended to the previous round.

Tested with import of the Product Key files : 
[Output_Product_Rev5.zip](https://github.com/microsoft/ConversationLearner-UI/files/3726643/Output_Product_Rev5.zip)

### DIFF
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -214,14 +214,24 @@ export class ObiDialogParser {
 
     // Generates TrainDialog instances from the dialog tree.
     private async getTrainDialogs(node: ObiDialogNode): Promise<CLM.TrainDialog[]> {
-        return this.getTrainDialogsIter(node, [])
+        return this.getTrainDialogsIter(node, [], node.intent)
     }
 
     // Recursive helper.
-    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[]):
+    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[],
+        intent: string | undefined):
         Promise<CLM.TrainDialog[]> {
         if (!node) {
             return []
+        }
+        // Intent may be carried forward from a previous node if that node did not create a TrainRound.
+        let currentIntent = intent
+        if (currentIntent) {
+            if (node.intent && node.intent !== currentIntent) {
+                throw Error(`Node intent ${node.intent} conflicts with incoming intent ${currentIntent}`)
+            }
+        } else {
+            currentIntent = node.intent
         }
         let rounds = [...currentRounds]
         // Build up a training round from any applicable steps in this node.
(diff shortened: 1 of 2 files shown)


src/Utils/obiDialogParser.ts
```diff
@@ -11,7 +11,8 @@ import * as stripJsonComments from 'strip-json-comments'
 
 enum OBIStepType {
     BEGIN_DIALOG = "Microsoft.BeginDialog",
-    END_TURN = "Microsoft.EndTurn",
+    END_DIALOG = "Microsoft.EndDialog",
+    END_TURN = "Microsoft.EndTurn",  // TODO(thpar) : Obsolete / delete ?
```
LarsLiden: I believe this is still a valid OBIStep type.  We just don't see it in any of the exported dialogs from CCI

src/Utils/obiDialogParser.ts
```diff
@@ -214,28 +215,47 @@ export class ObiDialogParser {
...
     // Generates TrainDialog instances from the dialog tree.
     private async getTrainDialogs(node: ObiDialogNode): Promise<CLM.TrainDialog[]> {
-        return this.getTrainDialogsIter(node, [])
+        return this.getTrainDialogsIter(node, [], node.intent)
     }
 
     // Recursive helper.
-    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[]):
+    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[],
+        intent: string | undefined):
```
mattmazzola: nit: would either put all the arguments on separate lines or all same line, not mixed

src/Utils/obiDialogParser.ts
tpmsr: Got it.  Removed comment.

src/Utils/obiDialogParser.ts

tpmsr:
Thanks for the quick reviews!

PR #1337, opened 2019-10-16
feat: allow validation tests to run in parallel

LarsLiden:


src/Utils/obiUtils.tsx
```diff
@@ -46,7 +46,7 @@ export async function getLogDialogActivities(
     conversationId: string | undefined,
     channelId: string | undefined,
     fetchLogDialogThunkAsync: (appId: string, logDialogId: string, replaceLocal: boolean, nullOnNotFound: boolean, noSpinner: boolean) => Promise<CLM.LogDialog>,
-    fetchActivitiesAsync: (appId: string, trainDialog: CLM.TrainDialog, userName: string, userId: string, useMarkdown: boolean) => Promise<CLM.TeachWithActivities>
+    fetchActivitiesAsync: (appId: string, trainDialog: CLM.TrainDialog, userName: string, userId: string, useMarkdown: boolean, noSpinner: boolean) => Promise<CLM.TeachWithActivities>
```
tpmsr: These could really use some documentation.  Please flesh out the function comments and add parameter docu.

LarsLiden: We've generally not been documenting parameters when they are self-explanatory, but they may just seem more self-explanatory to us as we've been working with the code base for a long time

mattmazzola: I don't think documentation fixes root of issue here. The functions like this should take objects to have `named`  parameters. The `, true, true, true)` is very bad as consumer doesn't know what each `true` means. Would be much better as

```
fetchLogDialogThunkAsync({
  appId,
  logDialogId,
  replaceLocal: true,
  nullOnNotFound: true,
  noSpinner: true
})
```

This is one of the reasons they added short-hand parameters to objects.

src/types/ActionObject.ts
```diff
@@ -196,17 +196,13 @@ export type FetchAction = {
     allTrainDialogs: CLM.TrainDialog[],
 } | {
     type: AT.FETCH_ACTIVITIES_ASYNC,
-    appId: string,
-    userName: string,
-    userId: string,
-    trainDialog: CLM.TrainDialog
+    noSpinner: boolean
```
mattmazzola: Does this mean `displaySpinner` or does `no` mean something else besides display?

src/routes/Apps/App/Testing.tsx
```diff
@@ -315,8 +295,50 @@ class Testing extends React.Component<Props, ComponentState> {
...
         if (this.state.testSet) {
             const testItems = this.state.testSet.getTestItems(sourceName)
-            await Util.setStateAsync(this, { testItems, testIndex: 0 })
-            await this.testNextTranscript()
+            await Util.setStateAsync(this, { testItems, completedTestIds: [], doneCount: 0 })
+            const testSlots: string[] = []
+            for (let i = 0; i < NUM_PARALLEL_TESTS; i = i + 1) {
+                testSlots.push(`ValidationTest ${i}`)
+            }
+            await Util.setStateAsync(this, {testSlots})
```
mattmazzola: nit: formatting

src/Utils/obiUtils.tsx
tpmsr: In this case, I think the thunk has enough params that it could use documentation.  I have no idea from looking at the signature why `userName` is required, or who it refers to; or why the `noSpinner` param is relevant for example.

src/routes/Apps/App/Testing.tsx
```diff
@@ -481,6 +503,11 @@ class Testing extends React.Component<Props, ComponentState> {
         return ''
     }
 
+    // Progress count is less jumpy when smoothed between in progress and completed
+    displayCount(): number {
+        return Math.round((this.state.doneCount + this.state.completedTestIds.length) * 0.5)
```
mattmazzola: I didn't understand what the "smoothing" part is.  If `doneCount` and `completedTestIds` are both positive integers the number would always be N + 0.5 which would up. Don't understand the need for `rouding`. Why not just `ceil` or nothing?

src/types/ActionObject.ts
LarsLiden: Means show no spinner.  Can rename it to "noSpinnerDisplay"

src/routes/Apps/App/Testing.tsx
LarsLiden: what formatting change are you suggesting?

src/Utils/obiUtils.tsx
LarsLiden: Right place to document this would be in logActions.   All our our actions have this issue....

tpmsr: I'm all for self-documenting code, but can also attest that as somebody new to the project, lack of actual documentation has been a **large** source of friction for getting spun up.

If you go with Matt's idea of using an object to store the params, that would make a convenient place to add documentation.

LarsLiden: Added comments to thunks

src/routes/Apps/App/Testing.tsx
LarsLiden: Done lags completed.  So display will immediately jump.  This averages the two and rounds to a whole number which smooths the display

src/types/ActionObject.ts
mattmazzola: Seems like a weird variable name to me. It allows double negatives and those should be avoided.

`displaySpinner: true` Display spinner
`displaySpinner: false` Don't display spinner
`noSpinnerDisplay: false` : ? Does that mean spinner is displayed because it was no display?

src/routes/Apps/App/Testing.tsx
mattmazzola: Did you try formatting it and seeing what changed? It's built into VSCode.
We do this so these changes don't keep showing up on other peoples reviews and distracting from the actual content of review.

You need spaces around curly braces
`await Util.setStateAsync(this, { testSlots })`

src/types/ActionObject.ts
LarsLiden: Generally I lean towards making the uncommon condition (the exception / added feature) to true.   but I'm happy to reverse it

src/routes/Apps/App/Testing.tsx
mattmazzola: I will look at enabling `prettier` again so we can avoid these conversations. They happening way too often. Last time we tried it seemed to change the code too much so it was turned off but saving the team from all this time would be more important.

src/Utils/obiUtils.tsx
LarsLiden: If we pass objects to the thunks we should change all of them at the same time.  It's weird to have one outlier

mattmazzola: Wow! Going from "I change things gradually" and intentionally introduce outliers to "we should change all of them" "It's weird to have one outlier". Great change, but I will believe it when I see it.

LarsLiden:
Does prettier have a config file that allows us to turn stuff on and off?  Didn’t like all the defaults

src/routes/Apps/App/Testing.tsx
mattmazzola: Oh, I see. I read `* 0.5` as `+ 0.5` which is why I was confused. I usually think of averages as dividing by total count of items in series such as `(x + y) / 2`.

src/Utils/obiUtils.tsx
tpmsr: Grazie!  For my own knowledge, though, who/what is the user referenced in the thunk params ?  I haven't yet seen us passing around users elsewhere in the code.

mattmazzola: The UI uses a simulated `user` object `{ name, id }` to recreate what happens when user is actually logged into something like Teams, Skype, etc.  All those channels would send this user object on the activity. Since we know this activity since it deterministic from the SDK the SDK can then check for it and know if it activity is coming from the administrative UI or some actual channel and treat it differently.

PR #1339, opened 2019-10-16
feat: Handle SwitchCondition nodes in .dialog import

tpmsr:
This PR adds initial support for `SwitchCondition` .dialog nodes.

* Branches from the `SwitchCondition` are traversed and included in construction of the dialog tree
* Conditions from each `SwitchCondition` case are used to generate enum entity values
  * Eg, if a switch has cases for `isProductKey == Windows` and `isProductKey == Office`, we will generate an enum entity `isProductKey`  with values `Windows` and `Office`:

enum_entity

Not implemented in this PR (coming later):
* Setting bot memory following the API call
* Adding "condition" restrictions to the actions generated from switch statements
* Generating turns for the input and output values of the API call
* Unit testing against interesting .dialog structures

Notes:
* Condition values are truncated to 10 characters during enum creation.  If multiple conditions map to the same truncated value, import will fail with an error message.
* There are some limitations on the contents of `SwitchCondition `nodes; see code comments for details

### DIFF
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -391,7 +391,6 @@ export class ObiDialogParser {
 
     /**
      * Creates enum entities and values for elements used in SwitchCondition comparisons, if they do not already exist.
-     * Noop for values that have already been created.
      * 
      * @param conditionalEntities dictionary key is the name of the value used in comparison (entity name);
      *     dictionary values are the distinct string values used across all comparisons of that entity.
--- a/src/Utils/obiDialogParser.ts
+++ b/src/Utils/obiDialogParser.ts
@@ -214,14 +214,24 @@ export class ObiDialogParser {
 
     // Generates TrainDialog instances from the dialog tree.
     private async getTrainDialogs(node: ObiDialogNode): Promise<CLM.TrainDialog[]> {
-        return this.getTrainDialogsIter(node, [])
+        return this.getTrainDialogsIter(node, [], node.intent)
     }
 
     // Recursive helper.
-    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[]):
+    private async getTrainDialogsIter(node: ObiDialogNode, currentRounds: CLM.TrainRound[],
+        intent: string | undefined):
         Promise<CLM.TrainDialog[]> {
         if (!node) {
             return []
+        }
+        // Intent may be carried forward from a previous node if that node did not create a TrainRound.
+        let currentIntent = intent
+        if (currentIntent) {
+            if (node.intent && node.intent !== currentIntent) {
+                throw Error(`Node intent ${node.intent} conflicts with incoming intent ${currentIntent}`)
+            }
+        } else {
+            currentIntent = node.intent
         }
         let rounds = [...currentRounds]
         // Build up a training round from any applicable steps in this node.
@@ -229,11 +239,12 @@ export class ObiDialogParser {
         if (obiDialog.steps) {
             let trainRound = await this.getTrainRoundfromOBIDialogSteps(obiDialog.steps)
             if (trainRound) {
-                if (node.intent) {
+                if (currentIntent) {
                     const extractorStep: CLM.TrainExtractorStep = {
-                        textVariations: this.getTextVariations(node.intent)
+                        textVariations: this.getTextVariations(currentIntent)
                     }
                     trainRound.extractorStep = extractorStep
+                    currentIntent = undefined  // Used the intent in this round, so reset it.
                 }
                 rounds.push(trainRound)
             }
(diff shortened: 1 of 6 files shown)


src/Utils/obiDialogParser.ts
```diff
@@ -78,7 +80,9 @@ export class ObiDialogParser {
         if (!mainDialog) {
             this.warnings.push(`Missing entry point. Expecting a .dialog file called "Entry.main"`)
         } else {
-            const rootNode = await this.collectDialogNodes(mainDialog)
+            let conditionalEntities: { [key: string]: Set<string> } = {}
```
mattmazzola: Seems like this should be `const`. I'm guessing you made it `let` because it was passed in to the `collectDialogNodes` but you likely didn't mean to indicate that the method will overwrite the `conditionEntities` to something like a new array. Yu likely meant that it would adjust the `Set`'s

src/Utils/obiDialogParser.ts
```diff
@@ -274,9 +302,9 @@ export class ObiDialogParser {
         return dialogs
     }
 
-    private async getTrainRoundfromOBIDialogSteps(steps: (string | OBITypes.OBIDialog)[]):
-        Promise<CLM.TrainRound | undefined> {
-        let trainRound: CLM.TrainRound | undefined
+    private async getScorerStepsFromOBIDialogSteps(steps: (string | OBITypes.OBIDialog)[]):
+        Promise<CLM.TrainScorerStep[]> {
+        let scorerSteps: CLM.TrainScorerStep[] = []
```
mattmazzola: seems like `const` again

src/Utils/obiDialogParser.ts
```diff
@@ -182,14 +189,14 @@ export class ObiDialogParser {
     /**
      * Collects dialog nodes from dialog-redirecting elements in the dialog `steps` section.
      */
-    private async collectDialogStepChildren(node: ObiDialogNode, steps: (string | OBITypes.OBIDialog)[]) {
+    private async collectDialogStepChildren(node: ObiDialogNode, steps: (string | OBITypes.OBIDialog)[],
+        conditionalEntities: { [key: string]: Set<string> }) {
```
mattmazzola: I've seen this is multiple reviews and didn't get answers on the other ones. Maybe better to ask why are you doing this?
Sometime it's an argument type other times it's a return type. Doesn't seem to be anything special about this one argument. I can't make sense of it other than something like an 80 character width limit rule.

LarsLiden: Generally like to see the return type defined

src/Utils/obiDialogParser.ts
```diff
@@ -369,6 +487,11 @@ export class ObiDialogParser {
             throw new Error('HTTP requests require url')
         }
         // TODO(thpar) : revisit logic for this.
+        // Note that we cannot do this 100% correctly in the current implementation, since actions (scorer steps)
+        // from a given dialog tree node $Y may be added to rounds from the previous dialog tree node $X if $Y does
+        // not have an extractor step, but we are calling this method during the handling of $X.
+        // To handle this 100% correctly, we'd need to do a multi-pass traversal of the dialog tree or build up a
+        // second tree-like representation of extractor and scorer steps.
```
LarsLiden: Is there any way to catch this if it happens?  Concerned about a silent failure.

src/Utils/obiDialogParser.ts

src/Utils/obiDialogParser.ts

src/Utils/obiDialogParser.ts
tpmsr: I have my editor split into 2 panes, so I'm generally just breaking up these long lines if they don't fit in the pane on ~half my monitor.

Updating to args on single lines per your earlier request.

src/Utils/obiDialogParser.ts
tpmsr: I think the import will fail if this happens, since we'd have a round with a terminal action followed by another scorer step.

src/Utils/obiDialogParser.ts
```diff
@@ -337,11 +386,80 @@ export class ObiDialogParser {
...
+            const truncated = value.substr(0, this.MAX_ENUM_VALUE_NAME_LENGTH)
+            if (!updatedValues[truncated]) {
+                updatedValues[truncated] = new Set([value])
+            } else {
+                // Some value with this truncated string already exists, throw an error if it's a new value.
+                // That would mean that we have 2 distinct condition values that map to the truncated string.
+                const fullValues = updatedValues[truncated]
+                if (!fullValues.has(value)) {
+                    const existing = fullValues.values().next().value
+                    throw new Error(`Can't create enum, values ${value} and ${existing} map to the same truncated string`)
```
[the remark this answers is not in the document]
tpmsr: I think I'd prefer to start with keeping it simple and see if we have a need for this.  I'd be concerned about generating ordinal names in that it's not going to be obvious to the user which entity was created for which `SwitchCondition`, and I think that might require us to persist the mapping somewhere as well.

src/Utils/obiDialogParser.ts
mattmazzola: It sounds like a problem with using word wrapping instead of actually putting a newline. This way you can satisfy personal preferences / conditions of small screen without influencing code expectations for everyone else.  Good to have separate lines for these especially long arguments names and types as opposed to something small like `index: number`

PR #1340, opened 2019-10-17
 feat: enable editing conditions from entities

mattmazzola:
Adds existing condition using the entity on to the entity modal.

image

Note: This new table is only shown for `number` LUIS type entites and not ENUMs even though they both can have conditions because when editing ENUM you can only change the value and since values are mutually exclusive this doesn't really make sense.

If you are renaming a value then this can already be done through the entity modal.

Basic flow is I have model with entity `myNumber` which is used as condition on action(s). I go to the `myNumber` entity modal and click edit on the condition. It opens the Condition modal where I can edit and save the changes.  It will then edit all the actions using this condition.  I can then close the entity modal since I technically only made changes to the actions, not to the entity.


src/components/modals/EntityCreatorEditor/EntityCreatorComponent.tsx
```diff
@@ -351,6 +386,14 @@ const Component: React.SFC<Props> = (props) => {
                 <FormattedMessageId id={FM.ENTITYCREATOREDITOR_DELETE_ENUM_ERROR_WARNING} />
             </div>}
         />
+        <ConditionCreatorModal
```
mattmazzola: I suppose I should rename this `ConditionModal` since it's not constrained to creating and should handle everything about conditions.

src/components/modals/EntityCreatorEditor/EntityCreatorContainer.tsx
```diff
@@ -128,62 +136,79 @@ class Container extends React.Component<Props, ComponentState> {
...
 
     UNSAFE_componentWillReceiveProps(nextProps: Props) {
-        if (nextProps.open !== this.props.open) {
-            // Build entity options based on current model locale
-            const currentAppLocale = nextProps.app.locale
-            const preBuiltLocale = PreBuiltEntities.find(entitiesList => entitiesList.locale === currentAppLocale)
-            if (!preBuiltLocale) {
-                throw new Error(`Could not find locale: ${currentAppLocale} within list of supported locales: ${PreBuiltEntities.map(e => e.locale).join(', ')}`)
-            }
+        if (nextProps.open) {
```
mattmazzola: Looks like a lot of changes but it was mostly indentation / white-space from splitting this `if` statement.  Previously this only allowed running when the modal was opened, but conditions have to be recomputed while the model is open.

cypress/integration/regressionTs/actionConditions.spec.ts
```diff
@@ -650,29 +679,105 @@ describe('Action Conditions', () => {
             })
         })
 
-        describe(`EnumConditions`, () => {
-            it(`when selecting different enum activities dependent on original enum should be marked with errors`, () => {
-                cy.get(s.trainDialogs.descriptions)
-                    .contains(testData.inputs.setEnumValue)
+        describe(`Condition Editing`, () => {
```
mattmazzola: Rest of changes are mostly whitespace, new test was for Condition Editing which this PR introduced

src/Utils/actionCondition.ts
```diff
@@ -140,3 +140,35 @@ export const getUniqueConditions = (actions: CLM.ActionBase[]): CLM.Condition[]
 
     return uniqueConditions
 }
+
+/**
+ * Returns actions using the existing condition with modifications of replaces it with the new condition.
```
tpmsr: "*and* replaces it" ?

src/components/modals/EntityCreatorEditor/EntityCreatorComponent.tsx
tpmsr: ConditionEditorModal ?

src/Utils/actionCondition.ts
mattmazzola: Hmm, yea, wording isn't right. I don't think `and` is better though, it makes it sound like it does A _and_ B, but it just returns actions. The comment was trying to describe the kind of modification the updated actions have. I think the issue was with the word "replaces" should be "replacing". 

Maybe this?
> Returns actions using the existing condition with modifications of existing condition replaced by new condition.

src/components/modals/EntityCreatorEditor/EntityCreatorComponent.tsx
mattmazzola: I don't think that's better, still restricts it to editing even though is primarily used for creating. I don't want people to think it's only for editing or only for creating and don't want to go down the path we have with other modals where we add CreatorEditor nonsense. I thought by dropping both terms it would imply it can do various things with Conditions

tpmsr: Maybe it's just me, but I consider creation and editing to be fungible.

Agree that "CreatorEditor" is not a good option.

src/Utils/actionCondition.ts
tpmsr: > Returns a copy of the input actions, which have `existingCondition` replaced with `newCondition` 

?

mattmazzola: Nah, it doesn't return copy of the input actions

tpmsr: Ah, that's actually a good argument for changing the method name to something like `updateActionConditions`.  It's not obvious from the name that it mutates the input args.

> Updates the input actions by replacing `existingCondition` with `newCondition` ?

mattmazzola: > it mutates the input args

It doesn't and we should avoid that.

tpmsr: Ok, I'm super confused.  I thought you just said it **doesn't** return a copy ?

mattmazzola: Correct, it only returns array of actions which had the existing condition not copy of entire input.

PR #1343, opened 2019-10-17
fix: upgrade Office Fabric (not affect picker dismiss anymore)

mattmazzola:
~Work around issue with Office Fabric~
~https://github.com/OfficeDev/office-ui-fabric-react/issues/10892~

~I also upgraded it to latest since I thought that was issue at first. Isn't needed but is good to be up to date anyways.~

Separated picker dismiss issues into other PR, this is just fabric upgrade now. Will likely hold until smoke tests are working unless  that takes long time.


mattmazzola:
Going to hold PR to try to isolate bugs before possibly introducing others.

PR #1344, opened 2019-10-18
fix: remove getTextFromItem

mattmazzola:
Split out the Fabric update PR https://github.com/microsoft/ConversationLearner-UI/pull/1343 as I want to reduce variables. The should prevent the modal from opening which was known bug breaking the tests. Hopefully we can get smoke tests passing before upgrading Office Fabric


mattmazzola:
Hmm, doesn't fix the tests, but fixes other issues so will merge anyways. I looked more into test failure and understand why now

PR #1346, opened 2019-10-18
Trying for a better build process

SDESkowronski:
- Run smoke and regression tests in parallel
- Base authoring key off of build number%5

### DIFF
--- a/.circleci/SetLuisAuthoringKey.js
+++ b/.circleci/SetLuisAuthoringKey.js
@@ -12,11 +12,13 @@ let buildNumber = +process.env.CIRCLE_BUILD_NUM
 // We have 5 LUIS Authoring Keys that we rotate through.
 // We use the Circle CI Build Number to help us get an index to each in sequence.
 // A Build + 2 Test Runs is 3 Workflows, each having its own Build Number.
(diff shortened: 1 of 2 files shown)


.circleci/SetLuisAuthoringKey.js
```diff
@@ -12,11 +12,13 @@ let buildNumber = +process.env.CIRCLE_BUILD_NUM
 // We have 5 LUIS Authoring Keys that we rotate through.
 // We use the Circle CI Build Number to help us get an index to each in sequence.
 // A Build + 2 Test Runs is 3 Workflows, each having its own Build Number.
-// The Build part does NOT consume a LUIS Authoring Keys. Both test runs use the same key.
-// Thus we need to use one LUIS Authoring Key for every third Build Number.
+// The Build part does NOT consume a LUIS Authoring Keys, which affects our algorithm,
+// thus we need to use one LUIS Authoring Key for every other Build Number.
```
mattmazzola: Minor, but you could not that build numbers aren't sequential.
People might things rotating through 5 keys is 0,1,2,3,4,1,2,3,4,0,1... but it doesn't work that way since build numbers can skip

SDESkowronski: Lets see if this helps before I get into all of that.

PR #1347, opened 2019-10-21
feat: [.dialog import] create required conditions for actions following SwitchCondition nodes

tpmsr:
`SwitchConditions` may have clauses like `productType == WINDOWS` and `productType == OFFICE`, where `productType` can be set as the result of an API call.

When we generate a `TrainDialog` from a dialog tree with this type of structure, we will create an enum entity called `productType` to store each potential returned value of the API call, in this case `WINDOWS` and `OFFICE`.

We would like to set the appropriate enum entity value as a required condition on the next action generated from the dialog tree. However, actions are created as a secondary pass after a `TrainDialog` is generated. Eg, there is an interactive or automated pass where each `TrainDialog` step is either matched to an existing action, or a new action is created.

This PR captures the `SwitchCondition` conditional statements and adds them to the appropriate `TrainScorerStep` so the conditions can be created when the actions are actually generated.

Notes :
* Conditions will only be created if "Auto action match" is selected during import
* This PR depends on https://github.com/microsoft/ConversationLearner-Models/pull/215


tpmsr:
In action:

condition-office

and

condition-windows

src/Utils/obiDialogParser.ts
```diff
@@ -456,9 +502,33 @@ export class ObiDialogParser {
...
         return newEntity
     }
 
+    /**
+     * Returns a `Condition` referencing an enum entity and enum value id, which is built from the entity and value name.
+     */
+    private getConditionFromRequiredEntity(requiredEntity: OBIUtils.ConditionEntityAndValue): CLM.Condition {
+        const conditionEntity = this.entities.find(e => e.entityName === requiredEntity.entity)
+        if (!conditionEntity) {
+            throw new Error(`Couldn't find entity ${requiredEntity.entity}`)  // Unexpected, shouldn't happen.
```
mattmazzola: nit: prefer comments on separate lines than code, or in other codes comments and code not on same lines. Also think the comment should be on the condition / `if` statement not the error throwing of the error if you're trying to say the condition occurring shouldn't happen not the throwing of the error.

src/Utils/obiDialogParser.ts
```diff
@@ -244,11 +254,26 @@ export class ObiDialogParser {
...
+     * @param requiredEntity will be set if entry to this node was gated by a `SwitchCondition`
+     *     step.  Similarly to `intent`, this value will be used to set a required entity if the
+     *     node generates a `TrainExtractorStep` and will otherwise be propagated down the tree.
+     */
     private async getTrainDialogsIter(
         node: ObiDialogNode,
         currentRounds: CLM.TrainRound[],
-        intent: string | undefined):
+        intent: string | undefined,
+        requiredEntity?: OBIUtils.ConditionEntityAndValue):
```
LarsLiden: Again.  Suggest diff var name

src/Utils/obiDialogParser.ts
tpmsr: Moved to separate lines.  I'm keeping the comment inside the `if` block though; I think it's more clear that the comment is related to that condition.  It looks strange to my eye to put the comment before the `if`.

src/Utils/obiDialogParser.ts

tpmsr:
Thanks for the reviews - updated and retested, will merge.

PR #1348, opened 2019-10-22
Fix Tests Broken by UI Changes

SDESkowronski:
The addition of adding Conditions to Actions caused test code to break. This should fix most if not all issues related to that.


cypress/support/Actions.js
```diff
@@ -37,6 +37,16 @@ export function CreateNewAction({
   if (uncheckWaitForResponse) actionModal.UncheckWaitForResponse()
 
   actionModal.SelectType(type)
+
+  // These are in a specific order so as to allow explicitly specifying these entities/conditions
+  // and for that to not be interfered with by the mechanisim in the UI that sets them automatically.
+  //
+  // You can still test the automatic setting of these fields by not explicitly setting the arguments
+  // of those that should be automatically set.
+  if (disqualifyingEntities) actionModal.SelectDisqualifyingEntities(disqualifyingEntities)
```
mattmazzola: Looks like a lot of formatting an re arranging changes. The actual change is here. Explicitly selecting the entity instead of typing the entity.

PR #1349, opened 2019-10-22
feat: add row selection to cypress label event

LarsLiden:
Cypress event tool can optionally choose row to select from.  Is backwards compatible:

Select "hello" in the 2nd row
```
     var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", index: 1 })
     event.initEvent("Test_SelectWord", true, true)}
     element.dispatchEvent(event)
```
or default to the first row
 ```
    var event = new CustomEvent("Test_SelectWord", { detail: "hello")
     event.initEvent("Test_SelectWord", true, true)}
     element.dispatchEvent(event)
```

### DIFF
--- a/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
+++ b/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
@@ -424,8 +424,8 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
 
     /**
      * For cypress end to end unit testing.  Event selects phrase in entity labeller
-     * The following will select the word "hello" in the second row (row is 1 based)
-     *   var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", row: 2 }})
+     * The following will select the word "hello" in the second row (index is 0 based)
+     *   var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", index: 1 }})
      *   event.initEvent("Test_SelectWord", true, true)}
      *   element.dispatchEvent(event)
      * 
@@ -435,14 +435,14 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
      *   element.dispatchEvent(event)
      */
     @autobind
-    onTestSelectWord(val: { detail: { phrase: string, row: number } } | { detail: string }) {
+    onTestSelectWord(val: { detail: { phrase: string, index: number } } | { detail: string }) {
         
         if (!val.detail) {
             throw new Error("Test_SelectWord expecting detail phrase")
         }
 
         const phrase: string = (typeof val.detail !== "string") ? val.detail.phrase : val.detail
-        const row: number = (typeof val.detail !== "string") ? (val.detail.row - 1) : 0
+        const index: number = (typeof val.detail !== "string") ? val.detail.index : 0
 
         const words = phrase.split(" ")
         const firstWord = words[0]
@@ -450,11 +450,11 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
 
         // Get row
         const rows = Array.from(document.querySelectorAll('[data-testid="extractor-response-editor-entity-labeler"]'))
-        if (row > rows.length - 1) {
+        if (index > rows.length || index < 0) {
             throw new Error("Row index does not exist")
         }
 
-        const selectedRow = rows[row]
+        const selectedRow = rows[index]
         // Get start div
         const tokens = Array.from(selectedRow.querySelectorAll(".cl-token-node"))
         const firstWordToken = tokens.filter(element => element.children[0].textContent === firstWord)[0]
--- a/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
+++ b/src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
@@ -414,18 +414,18 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
 
     componentDidMount() {
         // For end 2 end unit testing.
-        document.addEventListener("Test_SelectWord", this.onTestSelectWord)
+        document.addEventListener("Test_SelectWord", this.onTestSelectWord as any)
     }
 
     componentWillUnmount() {
         // For end 2 end unit testing.
-        document.removeEventListener("Test_SelectWord", this.onTestSelectWord)
+        document.removeEventListener("Test_SelectWord", this.onTestSelectWord as any)
     }
 
     /**
      * For cypress end to end unit testing.  Event selects phrase in entity labeller
      * The following will select the word "hello" in the second row (row is 1 based)
-     *   var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", r:ow: 2 }})
+     *   var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", row: 2 }})
      *   event.initEvent("Test_SelectWord", true, true)}
      *   element.dispatchEvent(event)
      * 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx [hunk 1]
```diff
@@ -422,16 +422,40 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
         document.removeEventListener("Test_SelectWord", this.onTestSelectWord)
     }
 
-    // For end 2 end unit testing.
+    /**
+     * For cypress end to end unit testing.  Event selects phrase in entity labeller
+     * The following will select the word "hello" in the second row (row is 1 based)
+     *   var event = new CustomEvent("Test_SelectWord", { detail: { phrase: "hello", r:ow: 2 }})
```
mattmazzola: typo `row: 2`, would also consider the word `index` to be more precise and dropping the `-1` logic. This is for test code not for end users so it doesn't help

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
```diff
@@ -422,16 +422,40 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
...
+     */
     @autobind
     onTestSelectWord(val: any) {
-        const phrase: string = val.detail
+        if (!val.detail) {
+            throw new Error("Test_SelectWord expecting detail phrase")
+        }
+
+        const phrase: string = val.detail.phrase || val.detail
+        const row: number = val.detail.row - 1 || 0
```
mattmazzola: I don't think the ` || 0` is correct.  I think the `- 1` would coerce the first portion into a number regardless of `row` value so ` || 0` would never evaluate an you could still get negative numbers

I believe it would be better to use `Math.max(val.detail.row - 1, 0)`

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
```diff
@@ -422,16 +422,40 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
...
+
         const words = phrase.split(" ")
         const firstWord = words[0]
         const lastWord = words[words.length - 1]
 
+        // Get row
+        const rows = Array.from(document.querySelectorAll(".entity-labeler"))
```
mattmazzola: Move to constant and change the selector to `[data-testid="extractor-response-editor-entity-labeler"]`. It's better to avoid code that relies on CSS for presentation when possible.

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
```diff
@@ -422,16 +422,40 @@ class ExtractorResponseEditor extends React.Component<Props, State> {
...
+     *   event.initEvent("Test_SelectWord", true, true)}
+     *   element.dispatchEvent(event)
+     * 
+     * Or default to the first row:
+     *   var event = new CustomEvent("Test_SelectWord", { detail: "hello"})
+     *   event.initEvent("Test_SelectWord", true, true)}
+     *   element.dispatchEvent(event)
+     */
     @autobind
     onTestSelectWord(val: any) {
```
mattmazzola: Is there a way to add type information to get rid of the `any`?  I understand anyone could trigger this event with arbitrary data so we can't guarantee type would need run time checks as you have but after those checks it should narrow the type.

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: good catch.  Missed then when I updated it to being 1 based

LarsLiden: Should be:
    const row: number = val.detail.row ? (val.detail.row - 1) : 0

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: done

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: onTestSelectWord(val: { detail: { phrase: string, row: number } } | { detail: string }) {
        if (!val.detail) {
            throw new Error("Test_SelectWord expecting detail phrase")
        }

        const phrase: string = (typeof val.detail !== "string") ? val.detail.phrase : val.detail
        const row: number = (typeof val.detail !== "string") ? (val.detail.row - 1) : 0

LarsLiden: Only thing is then you need to do this:
document.addEventListener("Test_SelectWord", this.onTestSelectWord as any)

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: I called it row rather then index as it is 1 based

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
mattmazzola: Why does it have to be `any`? It looks like it could be `EventListenerOrEventListenerObject`

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
mattmazzola: Yea, i noticed and i'm saying since people don't know how you implemented it row as 0 based or row as 1 based, it was removing ambiguity of calling it index. Anyways doesn't matter

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
mattmazzola: Don't understand why this was marked as resolved while not using the `max` which is safer.

It seems `{ detail: '', row: -4 }` is valid in your checks. Meaning -4 - 1 is -5
Which would attempt to do `rows[-5]` instead of fallback to 0 as `row[0]`

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
mattmazzola: partially done, but ok

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
SDESkowronski: I'd prefer it to be zero based since only a programmer is going to use it and that is what we expect. And yes I noticed you called it row, but it could also be called rowIndex if you really think that makes a difference.

SDESkowronski: Looks like there is a typo in the comment:
r : ow: 2
The extra colon.

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: The goad here isn't to validate user input, it's to differentiate between the two data types

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: Tried that.  Throws a type error

src/components/ExtractorResponseEditor/ExtractorResponseEditor.tsx
LarsLiden: Ok - will switched to zero based

PR #1350, opened 2019-10-22
chore: chip away at the wall of warnings

tpmsr:
There are a huge amount of warnings that come up when building the UI.

This PR should have no functional changes; it does cut out a lot of the warning messages.


src/Utils/dispatchUtils.ts
```diff
@@ -198,6 +194,22 @@ function associateModelDialogsWithDispatchActionsAndClean(sourceModelPairs: Sour
     })
 }
 
+const generateDispatchDialog = (rounds: CLM.TrainRound[]) => (
```
mattmazzola: What is the warning that caused you to move this? These types of edits are difficult to review. Reviewer can't tell if it's moved or moved _and_ edited etc and try to avoid these operations unless it's following some standard. Lars was on a bit of moving all private methods below the `render` function on all React components recently because some warning said public before private which didn't make sense for components as you want to be able to scroll to bottom and find the JSX.

src/actions/actionActions.ts
```diff
@@ -70,7 +70,7 @@ export const editActionThunkAsync = (appId: string, action: ActionBase) => {
 
             // Fetch train dialogs if any train dialogs were impacted
             if (deleteEditResponse.trainDialogIds && deleteEditResponse.trainDialogIds.length > 0) {
-                dispatch(fetchAllTrainDialogsThunkAsync(appId))
+                await dispatch(fetchAllTrainDialogsThunkAsync(appId))
```
mattmazzola: I don't think this is what we want. Most of these `dispatch` calls are supposed to fire-forget style / non-awaited.

src/components/modals/ActionCreatorEditor.tsx
```diff
@@ -145,7 +144,7 @@ const conditionalEntityTags = (entities: CLM.EntityBase[], actions: CLM.ActionBa
         .map(e => convertEntityToConditionalTags(e))
         .reduce((a, b) => [...a, ...b], [])
 
-    const uniqueConditionTags = [
```
mattmazzola: I'm not sure which warning was telling you do this this but I'm skeptical as some of these changes which remove helpful information. Even more ironic because you're usually the one to complain about lack of documentation. I do prefer to avoid unnecessary variable names and use `return` when it's a simple object that is self explanatory. It can prevent an extra variable which can become stale as return type changes, but these names added semantic information. People reading the code have to understand, what the operation does now. When it was named "unique" it was clear it reduced it to unique tags and how it reduced it was abstracted away

src/components/modals/ConditionModal.tsx
```diff
@@ -134,13 +138,13 @@ const Component: React.FC<Props> = (props) => {
     const [showNumberValue, setShowNumberValue] = React.useState(true)
     const [numberValue, setNumberValue] = React.useState(0)
     const [enumValueOptions, setEnumValueOptions] = React.useState<EnumOption[]>([])
+    const [selectedEnumValueOption, setSelectedEnumValueOption] = React.useState<EnumOption>()
```
mattmazzola: Again, what rule caused this. Hooks are meant to be co-located so they can be moved as custom hook together in future.  Eg. all the enum functionality is grouped. Would change this back.

src/components/modals/ConditionModal.tsx
```diff
@@ -227,26 +231,26 @@ const Component: React.FC<Props> = (props) => {
             return
         }
 
-        const condition: CLM.Condition = {
+        const cond: CLM.Condition = {
```
mattmazzola: I can't imagine what warning this could have been from.  What is a `cond`? Would change it back

src/services/poller.ts
```diff
@@ -35,7 +35,7 @@ export class Poller {
     // private id = Symbol(Math.floor(Math.random() * 100))
     private polls: ActivePoll[] = []
     constructor(options: IPollerOptions) {
-        global.setInterval(async () => await this.poll(), options.interval)
```
mattmazzola: I can't remember, but did this `await` get added to fix one the `undefined` errors? I would check and see before removing

src/Utils/dispatchUtils.ts
LarsLiden: I don't feel strongly about having privates at the bottom.  We could change the tslib to remove this warning.

src/actions/actionActions.ts
LarsLiden: I believe if you put a "void" in front of that call it tells tslint that you intended not to await the promise so:

`void dispatch(fetchAllTrainDialogsThunkAsync(appId))`

src/actions/actionActions.ts
```diff
@@ -110,10 +110,10 @@ export const deleteActionThunkAsync = (appId: string, actionId: string, removeFr
 
             // Fetch train dialogs if any train dialogs were impacted
             if (deleteEditResponse.trainDialogIds && deleteEditResponse.trainDialogIds.length > 0) {
-                dispatch(fetchAllTrainDialogsThunkAsync(appId));
+                await dispatch(fetchAllTrainDialogsThunkAsync(appId));
```
LarsLiden: Also needs to not await

src/components/modals/ActionCreatorEditor.tsx
LarsLiden: I suggest a comment before the return statement for clarificaiton

src/components/modals/ConditionModal.tsx
```diff
@@ -227,26 +231,26 @@ const Component: React.FC<Props> = (props) => {
...
-            condition.value = numberValue
+            cond.value = numberValue
         }
         else if (selectedEnumValueOption) {
             // TODO: Fix enum types
-            condition.valueId = selectedEnumValueOption.data.enumValueId!
+            cond.valueId = selectedEnumValueOption.data.enumValueId!
         }
 
         return condition
```
LarsLiden: Return value looks to be incorrect.  Should the "cond" with the var name change

src/components/modals/ActionCreatorEditor.tsx
mattmazzola: If the comment is just going to re state what the variable was named then it serves no purpose. Might as well put the variable back. This is why it was explicitly named to begin with.

src/Utils/dispatchUtils.ts
tpmsr: This was to address the "use before declaration" warning.

src/components/modals/ActionCreatorEditor.tsx
tpmsr: The lint here is "unnecessary local variable".  In these cases, if the function-level comment and return type aren't providing enough information to support eliding the variable name on the returned value, then I think we should adjust the function comment.  Please LMK which instances you feel this is the case for.

src/components/modals/ConditionModal.tsx
tpmsr: This is the "use before definition" lint also.

src/components/modals/ActionCreatorEditor.tsx
LarsLiden: We should either (1) remove the tslint requirement or (2) add a descriptive comment.   I vote for (2)

src/components/modals/ConditionModal.tsx
tpmsr: This is to avoid shadowing another `condition` variable in the same scope @ https://github.com/microsoft/ConversationLearner-UI/blob/c47043ce77a090c907e5e84f7ee6f5dc35c33d25/src/components/modals/ConditionModal.tsx#L190

src/components/modals/ActionCreatorEditor.tsx
tpmsr: Cool, took a swag at (2) for this function.

src/components/modals/ConditionModal.tsx
tpmsr: Thanks, good catch!

src/actions/actionActions.ts
tpmsr: Lars' suggestion seems to mollify the linter!

src/actions/actionActions.ts
tpmsr: Removed await from all dispatches here, and other files as well.

src/services/poller.ts
tpmsr: Looks like the await has always been there : https://github.com/microsoft/ConversationLearner-UI/pull/619/files#diff-079185a02aae96b096f8166810b1a099R35

src/components/modals/ConditionModal.tsx
mattmazzola: Then make it `conditionFromState` as opposed to other `condition` from props. I think it is better to add information to differentiate then take away and rising confusion or ambiguity.  Also, think the variable name should indicate what is in in isolation without relying on surrounding context.

tpmsr: Done.  Note that the context here is a 10-line function, including whitespace.

tpmsr:
I'm breaking this out into smaller PRs

PR #1352, opened 2019-10-23
fix: show transcript id when show error

LarsLiden:


src/types/TestObjects.ts
```diff
@@ -496,7 +496,7 @@ export class TestSet {
     // Get sourceName for a transcript
     sourceName(transcript: BB.Activity[]): string {
         if (transcript.length === 0 || !transcript[0].channelId) {
-            throw new Error("Transcript does not have a channelId")
+            throw new Error(`Transcript ${this.conversationId(transcript)} does not have a channelId`)
         }
         return transcript[0].channelId
```
mattmazzola: Unrelated to review but curious what is the relationship between `sourceName` and `channelId`. Is source name our internal word for it? Is the channel like Skype, Messager, Teams or something different?

LarsLiden:
Note that “skype” would still be a valid value for us.  (There’s just no “CCI-V1” channel)

PR #1354, opened 2019-10-25
feat: update bot memory for imported TrainDialog rounds containing deterministic API calls

tpmsr:
.dialog files may contain "deterministic" API calls.  For this situation, the dialog tree looks as follows : 

* Make an API call (eg, `getProductType`)
* Switch on result of API call
  * if `windows`, do $X
  * if `office`, do $Y

Here, due to the structure of the `switch` node, we assume that the API will only return either "windows" or "office".

The dialog importer builds a conversation tree for each arm of the switch statement, to emulate the conversation following either of the 2 API results.

This PR sets the bot memory for each `TrainDialog` to reflect the appropriate API results :

bot memory set

(_Don't mind the card rendering error in this screenshot, there's some unexpected newline in the card's .json file in my bot._)

### DIFF
--- a/src/Utils/obiUtils.tsx
+++ b/src/Utils/obiUtils.tsx
@@ -647,13 +647,13 @@ function setMemoryStateForApiActionWithSwitch(
             // The next action is not a SwitchCondition-gated node.
             continue
         }
-        // Get conditions from the SwitchCondition-gated node.
-        const conditions = scorerStepConditions[nextScorerStep.importId]
-        const enumConditionData = getEnumConditionData(entities, conditions)
         // Set the logic result on the *current* scorer step.
         if (!scorerStep.logicResult) {
             scorerStep.logicResult = { logicValue: undefined, changedFilledEntities: [] }
         }
+        // Get conditions from the SwitchCondition-gated node.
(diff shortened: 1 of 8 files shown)


src/Utils/obiUtils.tsx
```diff
@@ -591,6 +597,114 @@ export async function createImportedActions(
...
+    }
+}
+
+/**
+ * Gets the enum entity and enum value associated with the given conditions.
+ * @throws if the entity referenced in any `Condition` is not a valid enum.
+ */
+function getEnumConditionData(entities: CLM.EntityBase[], conditions: CLM.Condition[]): EnumDataFromCondition[] {
+    const output: EnumDataFromCondition[] = []
+    for (const condition of conditions) {
```
mattmazzola: It seems like this loop either throws error or adds `EnumDataFromCondition` which is the same as a `map`.

Seems like this could be reduced to function that extract enum data from a single condition and then expose it's mapping nature

`conditions.map(c => getEnumConditionData(c, entities))`

tpmsr: JS patterns just not fully ingrained yet - updated

PR #1356, opened 2019-10-29
chore: Add integration test for .dialog import

tpmsr:
Runs import over a pared-down set of OBI .dialog files and verifies the structure of the resulting output.

It's somewhat complex to manually set up some of the edge cases handled by .dialog import, such as scorer steps that need to be appended to previous rounds, or intents (extractor steps) that need to be carried forward to later rounds; the .dialog files in this PR exhibit these features.


tpmsr:
Interestingly, this test is failing on Linux systems and succeeding on Windows.  This would typically point to some issue with paths, but the paths generated by the tests are absolute - see examples in this CircleCI run where I'd added some debug output : https://circleci.com/gh/microsoft/ConversationLearner-UI/4786

I disabled the test on non-Windows systems for now, I don't think it's high enough priority to burn a tonne of time on it.

package.json
```diff
@@ -38,6 +39,7 @@
     "immer": "3.1.3",
     "immutable": "^3.8.2",
     "inline-css": "^2.4.1",
+    "klaw-sync": "^6.0.0",
```
mattmazzola: Hmm, hesitant about addition of these packages but if it was needed then suppose it's fine we have enough already. 

1. Was this functionality not able to be achieved with `fs-extra`?
2. Did the async version not work?

src/Utils/obiDialogParser.test.ts
```diff
@@ -0,0 +1,202 @@
...
+            for (const key of Object.keys(importResults.conditions)) {
+                const conditions = importResults.conditions[key]
+                expect(conditions.length).toEqual(1)
+                const condition: CLM.Condition = conditions[0]
+                expect(condition.condition).toEqual(CLM.ConditionType.EQUAL)
+            }
+
+            expect(importResults.trainDialogs.length).toEqual(4)
+            // Validate 1st TrainDialog.
+            {
```
mattmazzola: Are these here to force local scoping or only for aesthetics? They don't look functional.

package.json
tpmsr: 1. fs-extra doesn't have a file tree walker, klaw was extracted from it : https://github.com/jprichardson/node-fs-extra#what-happened-to-walk-and-walksync

2. Async probably would work as well, but I didn't see a need for it here.  The code would be a little more convoluted with async, so I was going for comprehensibility over performance.

src/Utils/obiDialogParser.test.ts
tpmsr: A bit of both.  Without the extra scope, instead of `dialog` and `round0` we'd have to write something like `dialog0` and `dialog2_round0`.

src/Utils/obiDialogParser.test.ts
```diff
@@ -0,0 +1,202 @@
...
+                    outputAction.actionId = `${actions.length}`
+                    actions.push(outputAction)
+                    resolve(outputAction)
+                })
+            }
+            const entities: CLM.EntityBase[] = []
+            const fakeCreateEntityThunk = (appId: string, entity: CLM.EntityBase) => {
+                return new Promise<CLM.EntityBase>((resolve) => {
+                    // Update the entity id and return the entity.
+                    let outputEntity = Util.deepCopy(entity)
```
mattmazzola: Didn't notice this before. I've been trying to get people to follow convention of using `const` wherever possible so when `let` is used you know it was needed and the variable will be reassigned.  Similar to usage of `Map`s. I wasn't sure if you knew or not.

tpmsr: Yep, still doing that mental switch in my head from C++ to TS.  `const` in C++ doesn't let you mutate the variable.  Updated.

PR #1359, opened 2019-11-02
 test: update tests to match rest other .ts tests

mattmazzola:
Few of the older tests were written before TypeScript was supported and used old style. Updated them to match other TypeScript tests.


cypress/integration/regressionTs/entityConflicts.spec.ts
```diff
@@ -241,41 +276,101 @@ describe('Entity Conflicts', () => {
...
+                    .click()
 
                 cy.get(s.trainDialogs.descriptions)
                     .contains(testData.userInput1)
                     .click()
 
-                trainDialog.SelectChatTurnExactMatch(testData.userInput1)
-                trainDialog.RemoveEntityLabel(labeledWord1, testData.entityName)
-                trainDialog.LabelTextAsEntity(labeledWord2, testData.entityName)
-                trainDialog.ClickSubmitChangesButton()
```
LarsLiden: General comment for this code review.  If you decide to not employ the helpful functions, I'd add comments above each of the actions to explain what the action is doing.  i.e.  // Label Text As Entity

mattmazzola: Yea, in future I'll add comments for some of the multi step operations that are more obscure. 

> helpful functions

This assumes those functions are helpful.  They have nice names make the surface more appealing but going any deeper and you likely won't consider them helpful for any of the more valuable metrics like debuggability, composibility, documentation, etc.

PR #1364, opened 2019-11-08
refactor: batch 1 of warning fixes

tpmsr:
Pulling changes from https://github.com/microsoft/ConversationLearner-UI/pull/1350

That PR fails smoke test, so trying to isolate safe changes.


src/Utils/dispatchUtils.ts
```diff
@@ -23,6 +29,8 @@ export function generateDispatcherSource(
             return generateRandomMultiTransferDispatcherSource(sourceModelPairs, 20, 10)
         case DispatcherAlgorithmType.TestData:
             return generateTestData(sourceModelPairs, 20, 10)
+        default:
+            throw new Error(`Unhandled algorithm type ${algorithmType}`)
     }
 
     throw new Error(`Could not associate Dispatcher Algorithm Type with algorithm. You passed: ${algorithmType}`)
```
mattmazzola: I thought this error was a bit more informative but minor since this feature isn't used anyways. It's probably unreachable code now, I'm surprised that wasn't a lint error on it's own.

tpmsr: I can update it to a break, although that might be more error prone if the rest of the function changes.  LMK!

PR #1368, opened 2019-11-08
feat: use models package for mock data generation

LarsLiden:


src/Utils/dialogUtils.test.ts
```diff
@@ -259,12 +257,12 @@ describe('dialogUtils', () => {
         test('found', () => {
             const trainDialog1C = copyTrainDialog()
             let result = findMatchingTrainDialog(trainDialog1C, [trainDialog1, trainDialog2, trainDialog3])
-            expect(result?.trainDialogId).toEqual(trainDialog1.trainDialogId)
+            expect(result && result.trainDialogId).toEqual(trainDialog1.trainDialogId)
```
tpmsr: Was the previous new syntax not working here ?

LarsLiden: Merge issue.  Will fix

PR #1374, opened 2019-11-14
fix: small bugfixes for OBI import

tpmsr:
* Fix read of undefined property if `channelData.clData` is not set
* Skip entity extraction during .dialog import if there are no custom trained entities in the model


tpmsr:
> Could revert the `package-lock.json` changes, this PR doesn't change packages.

Done

PR #1378, opened 2019-11-18
feat: Log Ranking and Paging

LarsLiden:
- LogRanker: gives each log a score for probability of having an error
- TreeUtil: moved tree building from Tree component into util so can share
- Log Fetching: now uses paging, store deleted logs locally in case of call failure
- Review page for seeing suspect logs


src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.  
+ * Licensed under the MIT License.
+ */
+import * as CLM from '@conversationlearner/models'
+import { TreeUtils } from './TreeUtils'
+
+export interface LogScore {
```
tpmsr: Please fix: these fields could benefit from some comments.  For example, "knownPath" is comprehensible to people who have been in design discussions on this, but probably obtuse to somebody new to the project.

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    private entities: CLM.EntityBase[]
+    private logDialogs: CLM.LogDialog[]
+    private trainDialogs: CLM.TrainDialog[]
+    private logScores: LogScore[]
+
+    constructor(logDialogs: CLM.LogDialog[], trainDialogs: CLM.TrainDialog[], entities: CLM.EntityBase[]) {
+        this.entities = entities
+        this.logDialogs = logDialogs
+        this.trainDialogs = trainDialogs
```
tpmsr: Why are these member variables rather than parameters to analyze() ?

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    analyze(): LogScore[] {
```
tpmsr: Please add comments.  Eg, returns a LogScore for each input LogDialog

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    // Return number scorer steps in TrainDialog with low confidence
+    private numberLowConfidence(logDialog: CLM.LogDialog, knownUtterances: string[]): number {
+        let numLowConfidence = 0
+        logDialog.rounds.forEach(r =>
+            r.scorerSteps.forEach(ss => {
+                if (ss.predictionDetails.scoredActions.length >= 2) {
+                    const scoredActions = ss.predictionDetails.scoredActions.sort(sa => sa.score)
+                    const scoreDiff = scoredActions[0].score - scoredActions[1].score
+                    if (scoreDiff < 0.2) {
```
tpmsr: nit : make the 0.2 a parameter with a default value ?

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+        )
+        return utterances
+    }
+
+    // Remove punctuation and lowercase
+    private clean(text: string): string {
+        const regex = /[!"#$%&'ʼ()*+,-./:;<=>?@[\]^_`{|}~]/g;
+        return text.replace(regex, '').toLowerCase()
+    }
+}
```
tpmsr: nit : final newline

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+        const utterances: string[] = []
+        trainDialog.rounds.forEach(r =>
+            r.extractorStep.textVariations.forEach(t => utterances.push(this.clean(t.text)))
+        )
+        return utterances
+    }
+
+    // Remove punctuation and lowercase
+    private clean(text: string): string {
+        const regex = /[!"#$%&'ʼ()*+,-./:;<=>?@[\]^_`{|}~]/g;
```
tpmsr: Comment seems wrong re: lowercase.  Are you trying to strip out any non-alphanumeric data ?  If so, this would work better and be Unicode-aware

`const regex = /[^\p{L}]/gu`

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+        return text.replace(regex, '').toLowerCase()
```
tpmsr: Maybe replace with space instead of empty string ?

src/Utils/TreeUtils.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+        else if (!round1.scorerSteps && !round2.scorerSteps) {
+            return true
+        }
+        else {
+            return JSON.stringify(round1.scorerSteps) === JSON.stringify(round2.scorerSteps)
+        }
+    }
+
+    private findMatchingRound(parent: TreeNode, round: CLM.TrainRound, roundIndex: number, trainDialog: CLM.TrainDialog, testOnly: boolean): TreeNode | null {
```
tpmsr: Hang on - why does this method new to build a temp tree, and why is the input parent mutated at line 185 ?

It feels like you should be able to navigate the existing tree without mutations to find a node containing a given round.

src/routes/Apps/App/Index.tsx
```diff
@@ -47,8 +48,8 @@ class Index extends React.Component<Props, ComponentState> {
 
         const thunk1 = this.props.fetchBotInfoThunkAsync(this.props.browserId, app.appId)
         const thunk2 = this.props.setCurrentAppThunkAsync(this.props.user.id, app)
-        // Note: We load log dialogs in a separate call as eventually we want to page
-        const thunk3 = this.props.fetchAllLogDialogsThunkAsync(app, packageId)
+        // Fetch the first 100 dialogs
+        const thunk3 = this.props.fetchLogDialogsThunkAsync(app, packageId, true)
         const thunk4 = this.props.fetchAppSourceThunkAsync(app.appId, packageId)
```
tpmsr: Please give these more descriptive names than thunk1-4  :)

src/Utils/LogRanker.ts
LarsLiden: Switched it to just being a util function rather than a class

src/Utils/LogRanker.ts
LarsLiden: Not sure what you mean.  The function does lowercase too.   Just stripping punctuation

src/Utils/LogRanker.ts
LarsLiden: shouldn't matter,  just doing an exact string match between utterances

src/Utils/LogRanker.ts
tpmsr: Ah - I read the comment as "Remove punctuation and also remove lowercase characters".

Regex suggestion still applies.

src/Utils/TreeUtils.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        test('inTree', () => {
```
mattmazzola: Strings in `test` are supposed to explain the test. `inTree` is not appropriate. E.g. "given tree and dialog within tree should return true"

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+                        numLowConfidence = numLowConfidence + 1
+                    }
+                }
+            })
+        )
+        return numLowConfidence
+    }
+
+    // Return list of all unique utterances in TrainDialogs
+    private getAllUtterances(trainDialogs: CLM.TrainDialog[]): string[] {
```
mattmazzola: Add the word unique to name: `getAllUniqueUtterances`

src/Utils/TreeUtils.ts
LarsLiden: badly named variable.  Throws it away if it finds a match of adds it to existing tree if not.  Renamed variable and added comment for clarification

src/routes/Apps/App/Index.tsx
LarsLiden: renames

src/Utils/LogRanker.ts
mattmazzola: Would also say the method should be named what it does. If it removes punctuation and lowercase, name is `removePunctuationAndLowercase` so people don't wonder what `clean` means.

src/Utils/TreeUtils.test.ts
```diff
@@ -0,0 +1,32 @@
...
+            const logRanker = new TreeUtils([trainDialog1, trainDialog2, trainDialog3], [])
```
mattmazzola: It's a weird error that this is named `logRanker`. I assume it was intentional since I can't imagine how you would not notice unless really copy paste, but it's very weird to me to construct instance of `TreeUtils` and somehow get an instance of `logRanker`.

I would expect `const treeUtil = new TreeUtil()`

Then the methods `isInTree` would make more sense on a tree instance than a log ranker instance.

src/react-intl-messages.ts
```diff
@@ -815,7 +823,9 @@ export default {
         [FM.BUTTON_IMPORT]: 'Import',
         [FM.BUTTON_INFO]: 'Info',
         [FM.BUTTON_LOAD]: 'Load',
+        [FM.BUTTON_MORE_LOGS]: 'More Logs',
```
mattmazzola: I think the button should have the verb `Load` to explain what it does. Clicking it will load more.  More logs is still ambigous

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,130 @@
...
+    const utterances: string[] = []
+    trainDialog.rounds.forEach(r =>
+        r.extractorStep.textVariations.forEach(t => utterances.push(clean(t.text)))
+    )
+    return utterances
+}
+
+// Remove punctuation and lowercase
+function clean(text: string): string {
+    const regex = /[!"#$%&'ʼ()*+,-./:;<=>?@[\]^_`{|}~]/g;
```
tpmsr: Here's a page that lists the unicode character classes : https://www.compart.com/en/unicode/category

If you want the full character class (eg, such as "all punctuation") you can just use the first letter (eg, `\p{P}`) rather than the 2-letter code.

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,130 @@
...
+// Return list of all utternaces in a TrainDialog
+function getUtterances(trainDialog: CLM.TrainDialog): string[] {
+    const utterances: string[] = []
+    trainDialog.rounds.forEach(r =>
+        r.extractorStep.textVariations.forEach(t => utterances.push(clean(t.text)))
+    )
+    return utterances
+}
+
+// Remove punctuation and lowercase
```
tpmsr: Suggestion : `Removes most punctuation and lowercases the input string.`

src/reducers/logDialogsReducer.ts
```diff
@@ -7,41 +7,77 @@ import { AT } from '../types/ActionTypes'
...
             if (action.replaceLocal) {
-                const existingLogDialog = state.find(d => d.logDialogId === action.logDialog.logDialogId)
+                const existingLogDialog = state.logDialogs.find(d => d.logDialogId === action.logDialog.logDialogId)
                 if (existingLogDialog) {
-                    return replace(state, action.logDialog, a => a.logDialogId)
+                    replace(state.logDialogs, action.logDialog, a => a.logDialogId)
                 }
                 else {
-                    return [...state, action.logDialog]
+                    state.logDialogs = [...state.logDialogs, action.logDialog]
```
mattmazzola: Given this list is likely large using push here is more appropriate. `state.logDialogs.push(action.logDialog)`

src/reducers/logDialogsReducer.ts
```diff
@@ -7,41 +7,77 @@ import { AT } from '../types/ActionTypes'
...
-            // Delete log dialog optimistically.  Will reload train dialogs on failure
-            return state.filter(dialog => dialog.logDialogId !== action.logDialogId);
+            // Remove from pending item list
+            state.pendingDelete = state.pendingDelete.filter(ld => ld.logDialogId !== action.logDialogId)
+            return 
+        case AT.DELETE_LOG_DIALOG_REJECTED:
+            // Restore pending item, remove from pending list
+            const restoreItem = state.pendingDelete.filter(ld => ld.logDialogId === action.logDialogId)
+            state.pendingDelete = state.pendingDelete.filter(ld => ld.logDialogId !== action.logDialogId)
+            state.logDialogs = [...state.logDialogs, ...restoreItem]
```
mattmazzola: Similar to above would look at `push`

src/Utils/TreeUtils.ts
```diff
@@ -0,0 +1,205 @@
...
+    private addTrainDialog(tree: TreeNode, trainDialog: CLM.TrainDialog): void {
+        let parent: TreeNode | null = tree
+        for (let [roundIndex, round] of trainDialog.rounds.entries()) {
+            parent = this.findMatchingRound(parent, round, roundIndex, trainDialog, false)
+            if (!parent) {
+                return
+            }
+        }
+    }
+}
```
tpmsr: Nit : missing final newline on file

src/Utils/LogRanker.ts
```diff
@@ -0,0 +1,132 @@
...
+        this.trainDialogs = trainDialogs
+    }
+
+    analyze(): LogScore[] {
+
+        // Create tree from trainDialogs
+        const trainTree = new TreeUtils(this.trainDialogs, this.entities)
+        const leafActionIds = trainTree.getLeafActionIds()
+        const allKnownUtterances = this.getAllUtterances(this.trainDialogs)
+        this.logScores = []
```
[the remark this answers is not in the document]
LarsLiden: where?

src/Utils/TreeUtils.test.ts
```diff
@@ -0,0 +1,32 @@
...
... { "textVariations": [{ "text": "Install a new language", "labelEntities": [] }] }, "scorerSteps": [{ "input": { "filledEntities": [], "context": {}, "maskedActions": [] }, "labelAction": "43e4afcc-542b-4349-940b-8d3947acdeb0", { "input": { "filledEntities": [], "context": {}, "maskedActions": [] }, "labelAction": "721387b2-f183-4f90-bbc4-67ad66f39b2f" }] }, { "extractorStep": { "textVariations": [{ "text": "Change my keyboard language", "labelEntities": [] }] }, "scorerSteps": [{ "input": { "filledEntities": [], "context": {}, "maskedActions": [] }, "labelAction": "9bbfa48c-7b66-4ace-97eb-10ade3585535" }, { "input": { "filledEntities": [], "context": {}, "maskedActions": [] }, "labelAction": "d3eb7878-467c-49b1-9ab5-90df78dd434b" }] }, { "extractorStep": { "textVariations": [{ "text": "No, I need to add a new keyboard language", "labelEntities": [] }] }, "scorerSteps": [{ "input": { "filledEntities": [], "context": {}, "maskedActions": [] }, "labelAction": "0c919378-06a8-4867-a5ea-00f1cd2f8f57" }] }], "clientData": { "importHashes": [] }, "initialFilledEntities": [], "createdDateTime": "2019-10-11T23:24:23.718253+00:00", "lastModifiedDateTime": "2019-10-11T23:24:23+00:00" } as any
```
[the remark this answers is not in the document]
LarsLiden: added comment

src/Utils/TreeUtils.test.ts
LarsLiden: nice catch.  missed this on refactor

public/index.css
```diff
@@ -142,9 +142,13 @@ html {
...
 }
-
+.cl-icon[data-icon-name="Cancel"] {
+  color: var(--color-error) !important;
+  vertical-align: sub;
+  font-size: 1.2em;
+}
 .cl-icon[data-icon-name="CheckMark"] {
-  color: green;
+  color: green !important;
```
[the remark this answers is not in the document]
LarsLiden: All checkmarks were originally green, then they switched to blue.  Guess we need to decide which we want.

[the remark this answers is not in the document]
LarsLiden: removed for now

PR #1381, opened 2019-11-19
fix: bug 2249 pick invalid action when model untrained

LarsLiden:
Mostly moving functions out of ActionScorer into utils so can be used more broadly


src/Utils/dialogEditing.ts
```diff
@@ -61,11 +61,14 @@ export async function onInsertAction(
     // Find top scoring Action
     let insertedAction = DialogUtils.getBestAction(uiScoreResponse.scoreResponse, actions, canEndSession)
 
-    // None were qualified so pick the first (will show in UI as invalid)
+    // None were qualified so try to find a valid one in unscored action
```
tpmsr: nit : suggest moving this comment inside the `if` statement since it only applies if the condition is met.

src/Utils/dialogUtils.tsx
```diff
@@ -283,6 +284,131 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
     return best
 }
 
+// Return first unscored action that is available
+export function getBestUnscoredAction(scoreResponse: CLM.ScoreResponse, actions: CLM.ActionBase[], entities: CLM.EntityBase[], memories: CLM.Memory[]): CLM.ScoredAction | undefined {
+    if (scoreResponse.unscoredActions.length === 0) {
+        return undefined
+    }
+    // Return first available unscored action
```
tpmsr: For my own understanding, what's the rationale behind returning the first unscored action here ?

src/Utils/dialogUtils.tsx
```diff
@@ -283,6 +284,131 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
...
+            return scoredAction
+        }
+    }
+    // If none available return the first one
+    const unavailableScoredAction = { ...scoreResponse.unscoredActions[0], score: 1 }
+    delete unavailableScoredAction.reason
+    return unavailableScoredAction
+}
+
+// Check if entity is in memory and return it's name
```
tpmsr: nit : its (possessive)

it's => it is

src/Utils/dialogUtils.tsx
```diff
@@ -283,6 +284,131 @@ export function getBestAction(scoreResponse: CLM.ScoreResponse, allActions: CLM.
...
+export function entityInMemory(entityId: string, entities: CLM.EntityBase[], memories: CLM.Memory[]): { match: boolean, name: string } {
+    const entity = entities.find(e => e.entityId === entityId);
+
+    // If entity is null - there's a bug somewhere
+    if (!entity) {
+        return { match: false, name: 'ERROR' };
```
tpmsr: more appropriate to throw in this case ?

src/Utils/dialogEditing.ts
LarsLiden: My pref is for comment above if for easier reading, but I changed the comment to clarify:
 // If none were qualified, try to find a valid one in unscored action

src/Utils/dialogUtils.tsx
LarsLiden: No - still want to be able to render.  User may have deleted the entity.

src/Utils/dialogUtils.tsx
LarsLiden: We have no scores, so we don't know which one would be best.  Picking the first is as good as any other.  Added a comment

PR #1389, opened 2019-11-27
feat: ability to specify which action is used for reprompt" rather than "none of the above 

LarsLiden:
Added ability for user to pick a particular re-prompt action (rather than always re-prompting with the same action.


src/components/modals/ActionCreatorEditor/ActionCreatorEditor.tsx
```diff
@@ -293,7 +296,8 @@ interface ComponentState {
     slateValuesMap: SlateValueMap
     secondarySlateValuesMap: SlateValueMap
     isTerminal: boolean
-    reprompt: boolean
+    shouldReprompt: boolean
```
mattmazzola: It's good that you renamed the variable to better suite a boolean, but I would rather it follow the pattern of the other boolean variables with `is` prefix. Should seems better suited when it's followed by a verb when then implies the action object is doing the reprompting, but it's actually the server. The action simply has passive properties that are observed.  It's a similar reason for `isTerminal` rather than `shouldRescore`.

src/components/modals/ActionCreatorEditor/ActionCreatorEditor.tsx
```diff
@@ -695,7 +702,22 @@ class ActionCreatorEditor extends React.Component<Props, ComponentState> {
...
+
+    isSelfReprompt(repromptActionId: string | undefined): boolean {
+        return (repromptActionId === REPROMPT_SELF || (this.props.action !== null && repromptActionId === this.props.action.actionId))
+    }
+
+    @autobind
+    onChangeRepromptType() {
+        // Toggle between reprompt-self and reprompt other action
+        this.setState(prevState => ({
+            repromptActionId: this.isSelfReprompt(prevState.repromptActionId) ? undefined : REPROMPT_SELF,
```
mattmazzola: No need to recompute twice inside a function, can save to `isSelfReprompt`

src/components/modals/index.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.  
  * Licensed under the MIT License.
  */
-import ActionCreatorEditor from './ActionCreatorEditor'
+import ActionCreatorEditor from './ActionCreatorEditor/ActionCreatorEditor'
```
mattmazzola: I would see if you can avoid the repeating pathing. I thought you wouldn't need this since you added the default export.

mattmazzola:
This feature seems like it would be better named "the ability specify which action is used for reprompt" rather than "none of the above"

src/components/modals/ActionCreatorEditor/ActionCreatorEditor.tsx
LarsLiden: Not sure I understand your comment.  "isReprompt" seems misleading as the action isn't a reprompt action.  It's stating whether the action should reprompt if it gets an unexpected input

src/components/modals/index.ts
LarsLiden: Good catch.  (vscode auto change when file moved)

src/Utils/util.ts
```diff
@@ -5,9 +5,15 @@
...
 import * as IntlMessages from '../react-intl-messages'
 import * as Const from '../types/const'
-import { MessageValue } from 'react-intl'
+import * as OF from 'office-ui-fabric-react'
 import * as moment from 'moment'
 import * as stringify from 'fast-json-stable-stringify'
+import { MessageValue } from 'react-intl'
+
+// Used to add id for running UI tests
+export interface ChoiceGroupOptionWithTestId extends OF.IChoiceGroupOption {
```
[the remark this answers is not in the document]
LarsLiden: yeah.  We don't really have a good file for things like this in the /types folder.   models.ts is probably the closest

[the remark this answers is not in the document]
mattmazzola: > We don't really have a good file for things like this

Why isn't `models.ts` a good fit?

It is collection of shared models / interfaces and this is a shared interface.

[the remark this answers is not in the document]
LarsLiden: It's what I used.  Though i'd prefer types.ts to avoid confusion with the models package

src/components/modals/ActionCreatorEditor/ActionCreatorEditor.tsx
mattmazzola: > "isReprompt" seems misleading as the action isn't a reprompt action.

I'm not suggesting "isReprompt" should be used. I'm saying the problem is trying to make a verb style variable into a boolean by adding "should".  It should be a noun. Verbs are for methods not properties.

> It's stating whether the action should reprompt if it gets an unexpected input

This is the problem.  In your mind, you think the action is doing the reprompting.  In my mind this is not the correct way to describe this behavior because it bakes in all the assumptions about unrelated things. The focus should be on why it repromts, and the effect of detecting unexpected input. The result of unexpected input could be anything but we previously chose to reprompt.  Now, we're saying it could be any action (even an action that is not a prompt) which means it's not reprompting anymore. It could be to escalate to human, or to start the conversation over, or something else like "lets try that again".

Names like `hasUnexpectedInputAction` seems more appropriate.

LarsLiden: As the field on ActionBase is called "repromptActionId", it think it's less confusing to have them aligned

PR #1398, opened 2019-12-03
fix: bug 2382 expected entity and reprompt are mutually exclusive

LarsLiden:


mattmazzola:
What's the messaging for these features? 

I thought it was advised to use "reprompt" for actions representing question? So they could automatically reprompt.

And yet we also say to use expected entity for these actions representing questions where the user input is likely an answer.

It seems there guidance that both these features should be used in this sutation and yet they are not compatible.

LarsLiden:
Agree that it's potentially confusing and hopefully when we get context from LUIS this issue will go away.  The problem with the current implementation is that the behavior of Expected Entity is: "if nothing is labelled by LUIS, assume the result is the Expected Entity", whereas reprompt shows a reprompt action when the result is out of domain.   If Expected Entity is set nothing is out of domain.

mattmazzola:
Yes, I know why they are mutually exclusive. I believe I was one who opened the bug. I'm saying we should make it distinct when to use expected entity and when to use reprompt.

LarsLiden:
Where would you propose to add that messaging?   (If things align with LUIS we'll resolve this issue before launch.)

mattmazzola:
I'm not sure I understand why the location of the messaging matters more than the contents of the messaging, I would put it in the UI with the rest of the messaging.

LarsLiden:
What would you propose for the messaging?  I don't think there's a clear explanation we can give at the moment.

mattmazzola:
This is ridiculous.

> What's the messaging for these features?

That's what I asked you originally and then you ask me the same thing.  It sounds like "We don't have clear explanation of when to use each feature." and that is the answer. They're both recommended for the same scenario and yet not compatible. It's confusing and complex. Hopefully we use the new LUIS features soon to fix it.

LarsLiden:
I'm asking you because I don't think think there is a clear explanation we can give.  If you have one, let's add it, otherwise let's hold off until we have a better solution

PR #1404, opened 2019-12-05
fix: bug 2356

LarsLiden:
update train dialog validity flags on replay from train dialog list view

### DIFF
--- a/src/routes/Apps/App/TrainDialogs.tsx
+++ b/src/routes/Apps/App/TrainDialogs.tsx
@@ -1071,6 +1071,21 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
                 this.props.trainDialogReplayThunkAsync as any,
             )
 
+            // Get new activities to check for errors or warnings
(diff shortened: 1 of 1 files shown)


src/routes/Apps/App/TrainDialogs.tsx
```diff
@@ -1071,6 +1071,21 @@ class TrainDialogs extends React.Component<Props, ComponentState> {
                 this.props.trainDialogReplayThunkAsync as any,
             )
 
+            // Get new activities to check for errors or warnings
+            const teachWithActivities = ((await this.props.fetchActivitiesThunkAsync(this.props.app.appId, newTrainDialog, this.props.user.name, this.props.user.id) as any) as Promise<CLM.TeachWithActivities>)
+            const replayError = DialogUtils.getMostSevereReplayError((await teachWithActivities).activities)
```
mattmazzola: It looks like you have incorrect `await` statement due to wrong types above.  Either way two `await`s is weird and wouldn't recommend this syntax in inline `await`s

LarsLiden: Ugh.. Sloppy coding. thx

LarsLiden: Wow. It's vscode that stuck that in.  I typed "a", TAB for auto-complete, and....

vccode

PR #1408, opened 2019-12-10
Fix typo in the condition

msft-paulkim:


msft-paulkim:
https://github.com/microsoft/ConversationLearner-Models/pull/226
FYI, above PR needs to be pushed to make this PR has a green build.

